PR #10, opened 2018-07-06
Append trailing slash to resource if necessary

kalafut:


kalafut:
I agree and am rescinding this.

PR #12, opened 2018-07-19
Add support for VMSS

beezly:
Extends the azure auth plugin to check the identity of VMs in a Virtual Machine Scale Set as well as standalone VMs.

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -181,13 +181,14 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 		return nil
 	}
 
-	if subscriptionID == "" || resourceGroupName == "" || (vmName == "" && vmssName == "") {
-		return errors.New("subscription_id, resource_group_name, and either vm_name or vmss_name are required")
+	if subscriptionID == "" || resourceGroupName == "" {
+		return errors.New("subscription_id and resource_group_name are required")
 	}
 
 	var principalID, location *string
 
-	if vmssName != "" {
+	switch {
+	case vmssName != "":
 		client := b.provider.VMSSClient(subscriptionID)
 		vmss, err := client.Get(ctx, resourceGroupName, vmssName)
 		if err != nil {
@@ -203,7 +204,8 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 
 		principalID = vmss.Identity.PrincipalID
 		location = vmss.Location
-	} else {
+
+	case vmName != "":
 		client := b.provider.ComputeClient(subscriptionID)
 		vm, err := client.Get(ctx, resourceGroupName, vmName, compute.InstanceView)
 		if err != nil {
@@ -220,6 +222,8 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 
 		principalID = vm.Identity.PrincipalID
 		location = vm.Location
+	default:
+		return errors.New("either vm_name or vmss_name is required")
 	}
 
 	// Ensure the principal id for the VM matches the verified token OID
--- a/path_login.go
+++ b/path_login.go
@@ -188,7 +188,7 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 	var principalID, location *string
 
 	switch {
-	case vmssName != "" && vmName == "":
+	case vmssName != "":
 		client := b.provider.VMSSClient(subscriptionID)
 		vmss, err := client.Get(ctx, resourceGroupName, vmssName)
 		if err != nil {
@@ -210,7 +210,7 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 		principalID = vmss.Identity.PrincipalID
 		location = vmss.Location
 
-	case vmName != "" && vmssName == "":
+	case vmName != "":
 		client := b.provider.ComputeClient(subscriptionID)
 		vm, err := client.Get(ctx, resourceGroupName, vmName, compute.InstanceView)
 		if err != nil {
--- a/path_login.go
+++ b/path_login.go
@@ -188,7 +188,7 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 	var principalID, location *string
 
 	switch {
-	case vmssName != "":
+	case vmssName != "" && vmName == "":
 		client := b.provider.VMSSClient(subscriptionID)
 		vmss, err := client.Get(ctx, resourceGroupName, vmssName)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


chrishoffman:
Thanks for the contribution!  We have been looking at this internally and the problem is that you cannot tell from the instance metadata if the machine is part of a scale set.  The group placement sort of tells you but there is no explicit method that we have found.

This PR addresses this by adding a new parameter for a specific vmss machine name.  The issue is that the service integrating with Vault needs to know that it is part of a Scale Set.  This isn't a great experience because if you move your application from a VM to a Scale Set, you now have to change your integration.  There really isn't an alternative right now but we are in discussions with the Azure team about better ways to solve this.  

I'm going to keep this open for now but may close it if a better solution is found.

beezly:
@chrishoffman - thanks for the quick response! The Instance Metadata Service API versions `2017-08-01` and before didn't include `vmScaleSetName` in the response, but it looks like `2017-12-01` and above do, so it's relatively easy for a service to discover if it's running on a scale set or not.

chrishoffman:
Oh really?  Let me check again.  Thanks for the info!

chrishoffman:
For whatever reason, I was using `2017-08-01` in my testing.  I can confirm now that scale set name is included in the metadata service which is great news.

path_login.go
```diff
@@ -170,30 +175,58 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
...
+		return errors.New("subscription_id, resource_group_name, and either vm_name or vmss_name are required")
 	}
 
-	client := b.provider.ComputeClient(subscriptionID)
-	vm, err := client.Get(ctx, resourceGroupName, vmName, compute.InstanceView)
-	if err != nil {
-		return errwrap.Wrapf("unable to retrieve virtual machine metadata: {{err}}", err)
+	var principalID, location *string
+
+	if vmssName != "" {
```
chrishoffman: Can you change this to a switch statement and have the default of missing both values return an error?

path_login.go
```diff
@@ -170,30 +175,58 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
...
+		client := b.provider.VMSSClient(subscriptionID)
+		vmss, err := client.Get(ctx, resourceGroupName, vmssName)
+
+		if err != nil {
+			return errwrap.Wrapf("unable to retries virtual machine scale set metadata: {{err}}", err)
+		}
+
+		if vmss.Identity == nil {
```
chrishoffman: I haven't checked but do vms and vmss share the same identity type?  If so, you can just pull out the identity and share the logic between these two methods.

path_login.go
```diff
@@ -170,30 +175,58 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
...
-	client := b.provider.ComputeClient(subscriptionID)
-	vm, err := client.Get(ctx, resourceGroupName, vmName, compute.InstanceView)
-	if err != nil {
-		return errwrap.Wrapf("unable to retrieve virtual machine metadata: {{err}}", err)
+	var principalID, location *string
+
+	if vmssName != "" {
+		client := b.provider.VMSSClient(subscriptionID)
+		vmss, err := client.Get(ctx, resourceGroupName, vmssName)
+
```
chrishoffman: Small style thing but can you "attach" the error handling to the function that is returning the error by removing the extra newline?  This happens a couple of times in this PR.

path_login.go
beezly: Sadly they don't. One is a VirtualMachineIdentity and the other is VirtualMachineScaleSetIdentity - although (in the current SDK) they do have identical contents.

chrishoffman: No problem. This works.

beezly:
@chrishoffman updated with the suggestions you made. I've added a `bound_scale_sets` option which behaves much like the other `bound_*` options along with some tests. If bound_scale_sets is defined and a vm which _isn't_ in a VMSS attempts to use the role, then this code will reject it with a (hopefully) sensible error!

beezly:
If it's useful, I've been testing this out with a vault instance on Azure using this gist: https://gist.github.com/beezly/63aae866aa2219a05dfae52c5171b2ee

path_login.go
```diff
@@ -37,6 +37,10 @@ func pathLogin(b *azureAuthBackend) *framework.Path {
 				Type:        framework.TypeString,
 				Description: `The name of the virtual machine.`,
 			},
+			"vmss_name": &framework.FieldSchema{
+				Type:        framework.TypeString,
+				Description: `The name of the virtual machine scale set the instance is in`,
```
kalafut: nit: end the description with: `.`

path_login.go
```diff
@@ -170,30 +175,69 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
...
-	if err != nil {
-		return errwrap.Wrapf("unable to retrieve virtual machine metadata: {{err}}", err)
+	var principalID, location *string
+
+	switch {
+	case vmssName != "":
+		client := b.provider.VMSSClient(subscriptionID)
+		vmss, err := client.Get(ctx, resourceGroupName, vmssName)
+		if err != nil {
+			return errwrap.Wrapf("unable to retries virtual machine scale set metadata: {{err}}", err)
```
kalafut: typo: retries -> retrieve

path_login.go
```diff
@@ -170,30 +175,69 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
...
+
+		// Check bound scale sets
+		if len(role.BoundScaleSets) > 0 {
+			return errors.New("bound scale set defined but this vm isn't in a scale set")
+		}
+
+		principalID = vm.Identity.PrincipalID
+		location = vm.Location
+	default:
+		return errors.New("either vm_name or vmss_name is required")
```
kalafut: @chrishoffman Do you think providing _both_ names should also be considered an error?

chrishoffman: Agreed.  I had that comment at one point but it appears github ate it.

path_login.go
```diff
@@ -170,30 +175,69 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
...
 	}
 
-	client := b.provider.ComputeClient(subscriptionID)
-	vm, err := client.Get(ctx, resourceGroupName, vmName, compute.InstanceView)
-	if err != nil {
-		return errwrap.Wrapf("unable to retrieve virtual machine metadata: {{err}}", err)
+	var principalID, location *string
+
+	switch {
+	case vmssName != "" && vmName == "":
```
chrishoffman: Instead of this, would you mind just adding a switch at the top when both are not empty and error on that case.  I think that would be a little more clear.

path_login.go
beezly: @kalafut @chrishoffman thanks for the feedback. I've just added a check for that.

path_login.go
beezly: Will do!

beezly:
@chrishoffman that last commit also adds a test to ensure that it does fail if both `vm_name` and `vmss_name` are supplied.

path_login.go
beezly: @chrishoffman I've just realised I might have misunderstood your intent here. Are you saying the test to check if both vmss_name and vm_name are set should go at the top of `verifyResource` rather than at the top of _this_ switch statement?

chrishoffman: Nope, this looks good to me.

chrishoffman:
Thanks again for this! 💯

PR #15, opened 2018-11-10
Cache Azure authorizer

kalafut:


azure.go
```diff
@@ -136,6 +142,23 @@ func (p *azureProvider) VMSSClient(subscriptionID string) (vmssClient, error) {
 }
 
 func (p *azureProvider) getAuthorizer() (autorest.Authorizer, error) {
+	p.lock.RLock()
+	unlockFunc := p.lock.RUnlock
+	defer func() { unlockFunc() }()
```
ncabatoff: I don't think this is worth doing, I would just release the read lock inside the if block before returning.  It's the very next line after all.

azure.go
```diff
@@ -136,6 +142,23 @@ func (p *azureProvider) VMSSClient(subscriptionID string) (vmssClient, error) {
...
+	if p.authorizer != nil && time.Now().Before(p.authorizerExpiration) {
+		return p.authorizer, nil
+	}
+
+	// Upgrade lock
+	p.lock.RUnlock()
+	p.lock.Lock()
+	unlockFunc = p.lock.Unlock
+
+	if p.authorizer != nil && time.Now().Before(p.authorizerExpiration) {
```
ncabatoff: Why are we doing this a second time?

azure.go
```diff
@@ -158,6 +181,8 @@ func (p *azureProvider) getAuthorizer() (autorest.Authorizer, error) {
 			return nil, err
 		}
 	}
+	p.authorizer = authorizer
```
ncabatoff: Can we wait until here to grab the write lock?

azure.go
ncabatoff: Or rather, RLock, fetch p.authorizorExpiration, RUnlock, then if/return/etc.

kalafut: This is a pattern we've used a few places where an upgrade is necessary. I've carried it forward mainly b/c of precedent, but I do like getting the deferred unlock out of the way early so there is no question later (more an issue for a longer function that might pick up another return).

PR #23, opened 2019-07-02
Tokenutilize Azure auth

jefferai:


path_role.go
```diff
@@ -10,12 +10,13 @@ import (
...
 	"github.com/hashicorp/vault/sdk/helper/policyutil"
+	"github.com/hashicorp/vault/sdk/helper/tokenutil"
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
 // pathsRole returns the path configurations for the CRUD operations on roles
 func pathsRole(b *azureAuthBackend) []*framework.Path {
-	return []*framework.Path{
+	p := []*framework.Path{
 		&framework.Path{
```
kalafut: While we're here, consider removing the reduntant `&framework.Path`s and `&framework.FieldSchema`s.

path_role.go
```diff
@@ -243,39 +273,99 @@ func (b *azureAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logica
...
-	if policiesRaw, ok := data.GetOk("policies"); ok {
-		role.Policies = policyutil.ParsePolicies(policiesRaw)
-	}
+	if err := role.ParseTokenFields(req, data); err != nil {
+		return logical.ErrorResponse(err.Error()), logical.ErrInvalidRequest
+	}
+
+	// Handle upgrade cases
+	{
+		policiesRaw, ok := data.GetOk("token_policies")
```
kalafut: In all of these cases, the result term (e.g. `policiesRaw` here) isn't used, but the variable name _is_ used in inner blocks. Probably clearer to use `_` for the top level `GetOK`.

path_role.go
```diff
@@ -243,39 +273,99 @@ func (b *azureAuthBackend) pathRoleCreateUpdate(ctx context.Context, req *logica
...
+			}
+		}
 	}
 
-	if tokenTTLRaw, ok := data.GetOk("ttl"); ok {
-		role.TTL = time.Duration(tokenTTLRaw.(int)) * time.Second
-	} else if req.Operation == logical.CreateOperation {
-		role.TTL = time.Duration(data.Get("ttl").(int)) * time.Second
+	if role.TokenPeriod > b.System().MaxLeaseTTL() {
+		return logical.ErrorResponse(fmt.Sprintf("token period of '%q' is greater than the backend's maximum lease TTL of '%q'", role.TokenPeriod.String(), b.System().MaxLeaseTTL().String())), nil
```
kalafut:
```suggestion
		return logical.ErrorResponse("token period of %q is greater than the backend's maximum lease TTL of %q", role.TokenPeriod.String(), b.System().MaxLeaseTTL().String()), nil
```

jefferai: ErrorResponse takes in a bare string...

kalafut: It can also take a format string plus  parameters. And %q already adds wrapping quotes, so the extras were removed.

jefferai: ...since when? :-)

jefferai: I don't really know why it had quotes plus %q and didn't really look, I just kept the error as is and just adjusted the parameter

Issue #26, opened 2019-12-06
"SubscriptionNotFound" error when client tries to authenticate in Azure Gov


null-route:
Hi all,

After a day of troubleshooting I still haven't been able to make any progress on this and I'm beginning to think it may not be a simple misconfiguration. Is anyone successfully using the Vault Azure authentication backend in Azure Gov? Here is my current configuration (Vault servers are running 1.2.3):

- An application registration has been created, per https://www.vaultproject.io/docs/auth/azure.html#prerequisites-, and given the documented vm/vmss read permissions.
- An Azure auth backend has been configured within Vault and provided with the application registration client_id and client_secret that were created previously. Config snippet:
```
[server]$ vault read /auth/azure/config
Key            Value
---            -----
client_id      $vault_app_reg_client_id
environment    AzureUSGovernmentCloud
resource       https://management.usgovcloudapi.net/
tenant_id      $tenant_id
```
When an Azure Gov client VM attempts to authenticate using this authentication backend, they receive the following error (request parameters included):
```
[client]$ curl --request POST --data '{"role": "$CLIENT_ROLE", "jwt": "$CLIENT_ACCESS_TOKEN", "resource_group_name": "$CLIENT_RG", "vm_name": "$CLIENT_VM_NAME", "subscription_id": "$CLIENT_SUBID"}' https://$VAULT_IP:8200/v1/auth/azure/login -s | jq
{
  "errors": [
    "unable to retrieve virtual machine metadata: compute.VirtualMachinesClient#Get: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code=\"SubscriptionNotFound\" Message=\"The subscription '$CLIENT_SUBID' could not be found.\""
  ]
}
```

However, the Vault application registration credentials that were loaded in the backend can both read subscriptions and VMs, so they have the required permissions. The following commands can be used to demonstrate:
```
[server]$ export ACCESS_TOKEN="$(curl -X POST -d 'grant_type=client_credentials&client_id=$VAULT_CLIENT_ID&client_secret=$VAULT_CLIENT_SECRET&resource=https://management.usgovcloudapi.net/' https://login.microsoftonline.com/$TENANT_ID/oauth2/token -s | jq -r '.access_token')"
[...]

PR #27, opened 2020-01-07
Fix ComputeClient and VMSSClient to use correct baseURL

thardie:
ComputeClient and VMSSClient didn't pass BaseURL to the underlying
call, so if you were running in anything other than Azure public cloud,
the wrong endpoint would be queried, leading to a 404.

Fixes https://github.com/hashicorp/vault-plugin-auth-azure/issues/26

### DIFF
--- a/azure.go
+++ b/azure.go
@@ -121,7 +121,7 @@ func (p *azureProvider) ComputeClient(subscriptionID string) (computeClient, err
 		return nil, err
 	}
 
-	client := compute.NewVirtualMachinesClient(subscriptionID)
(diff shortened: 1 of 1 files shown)


kalafut:
@thardie Thanks for the contribution! I've not tested it yet but it looks reasonable.

The CLA is showing as unsigned. If you did sign it, check the emails on the commit and your github account, as they need to match.

thardie:
> @thardie Thanks for the contribution! I've not tested it yet but it looks reasonable.
> 
> The CLA is showing as unsigned. If you did sign it, check the emails on the commit and your github account, as they need to match.

I did see that. I need our legal department to give a thumbs up, then I'll sign it.

thardie:
CLA signed.

PR #40, opened 2021-02-19
Add GUID identifier

catsby:
# Overview

Adds a GUID into the user-agent header, depending on the Vault version (enterprise or not)

This is a companion PR to https://github.com/hashicorp/vault-plugin-secrets-azure/pull/55 in the Azure secrets plugin

### DIFF
--- a/util.go
+++ b/util.go
@@ -3,12 +3,10 @@ package azureauth
 import (
 	"encoding/json"
 	"fmt"
-	"os"
-	"runtime"
 	"strings"
 	"time"
 
-	"github.com/hashicorp/vault/sdk/helper/pluginutil"
+	"github.com/hashicorp/vault/sdk/helper/useragent"
 	"github.com/hashicorp/vault/sdk/version"
 )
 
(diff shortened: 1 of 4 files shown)


util.go
```diff
@@ -46,12 +47,31 @@ func strListContains(haystack []string, needle string) bool {
...
-	version := os.Getenv(pluginutil.PluginVaultVersionEnv)
+	pluginVersion := os.Getenv(pluginutil.PluginVaultVersionEnv)
+
 	projectURL := "https://www.vaultproject.io/"
 	rt := runtime.Version()
-	return fmt.Sprintf("Vault/%s (+%s; %s)", version, projectURL, rt)
+
+	// assign a GUID to the user-agent used in making requests
+	// create the basic user agent string
+	ua := fmt.Sprintf("Vault/%s (+%s; %s)", pluginVersion, projectURL, rt)
```
calvn, line 64: How come we're setting the user agent in here slightly differently (i.e. not calling `useragent.String()` from `sdk/helper/useragent`)? `pluginVersion` in here can be an empty string if ran as a builtin plugin. Is there a particular reason we're constructing the string manually vs calling the helpers on the `useragent` package similar to:

https://github.com/hashicorp/vault-plugin-secrets-azure/blob/8e44df36e43f6e0bfb99e0ffa16d5b90e1150127/provider.go#L83-L88

catsby: I've updated this to use `sdk/helper/useragent`

calvn (approved):

PR #39, opened 2020-11-29
Go1.15 & misc changes

elsesiy:
# Overview
This PR doesn't make any functional changes but instead focuses on upkeep & code quality.

## Changes
Update to go1.15
Update all dependencies
Remove ping for Azure IMDS during provider setup since the existing code is not error checked & the value-add is unclear
Replace deprecated 'Callbacks' with 'Operations' API
Use test.Helper() where applicable
Remove superfluous vendor exclusion for go list
Remove superfluous import alias for go-oidc & cleanhttp
Explicitly ignore unused func args via _
Remove unnecessary type definitions for inlined framework.FieldSchema usages
Remove invalid build for darwin/386

# Contributor Checklist
[X] Backwards compatible

PR #43, opened 2021-03-04
mod: pin x/oauth2 to bf48bf16ab8d

calvn:
# Overview
Pinning x/oauth2 to a version that doesn't depend on google.golang.org/grpc v1.30.0 or higher. This is to avoid incompatibility issues with the etcd module whenever both modules are imported by the same parent (e.g. Vault) 

# Design of Change
```
go get golang.org/x/oauth2@bf48bf16ab8d
go mod vendor
go mod tidy
```

- https://github.com/etcd-io/etcd/issues/12124
- #42 
- #39


HridoyRoy (approved):
👍 -- even though it's a rollback, it allows us to be compatible with all versions of grpc, which I think is worth the risk of pinning x/oauth2 to a particular version.

PR #55, opened 2022-06-29
Switch to Common Release Tooling

swenson:
# Overview

* Remove CircleCI integration
* Add GitHub actions for testing
* Add GitHub actions to cross-compile on every platform

I haven't tested releasing yet, but artifacts seemed to be uploading in
the proper place for them to be picked up by CRT.

# Design of Change

N/A

# Contributor Checklist
* [x] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet (N/A)
* [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
* [x] Backwards compatible


.github/workflows/build.yaml
```diff
@@ -0,0 +1,138 @@
...
+          - os: darwin
+            arch: "386"
+          - os: darwin
+            arch: arm
+          - os: freebsd
+            arch: arm64
+          - os: netbsd # I guess "Of course it runs NetBSD" is no longer true
+            arch: arm64
+          - os: openbsd
+            arch: arm
```
tvoran, line 94: Looks like vault is building this combo, so we might want to include it for the plugins? https://github.com/hashicorp/vault/actions/runs/2586375558

swenson: Hmm. I tried to add it and got:

```
No valid platforms to build for. If you specified a value
for the 'os', 'arch', or 'osarch' flags, make sure you're
using a valid value.
```

Let me see how Vault main gets a build of openbsd arm.

tvoran: Yeah from what I recall, CRT builds don't use/need gox since each os/arch combo is built in a separate job.

swenson: We could switch to using `GOOS`, etc., if I don't get any traction about updating `gox`

We still use separate builds for performance (it can take several minutes for 1 job to build every platform we require).

swenson:
@tvoran this is ready for review now.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,135 @@
...
+          ([ -d out ] || mkdir out) &&
+          zip -r -j out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.os }}_${{ matrix.arch }}.zip bin/*
+
+      - name: Test that it loads with Vault
+        if: matrix.os == 'linux' && matrix.arch == 'amd64'
+        run: |
+          docker pull vault &&
+          docker run -d --name vault --mount type=bind,src=${PWD}/bin,dst=/vault/file --cap-add=IPC_LOCK -e VAULT_ADDR='http://127.0.0.1:8200' -e 'VAULT_TOKEN=myroot' -e 'VAULT_DEV_ROOT_TOKEN_ID=myroot' -e 'VAULT_DEV_LISTEN_ADDRESS=127.0.0.1:8200' vault server -dev -dev-plugin-dir /vault/file &&
+          timeout 30 sh -c 'until docker exec -t vault vault status; do sleep 1; done' &&
+          docker exec -t vault vault auth enable vault-plugin-auth-azure
```
tvoran, line 129: Nice!

tvoran (approved):
Just a couple minor comments.

.release/ci.hcl
```diff
@@ -0,0 +1,130 @@
...
+    repository = "crt-workflows-common"
+    workflow = "security-scan-binaries"
+    config = "security-scan.hcl"
+  }
+
+  notification {
+    on = "fail"
+  }
+}
+
```
tvoran, line 66: Since we're producing darwin and windows binaries, do we need to notarize them like vault does? https://github.com/hashicorp/vault/blob/main/.release/ci.hcl#L90-L140

swenson: Good catch, thanks!

swenson:
Thanks. I'm going to go ahead and merge this to try to test the release pipeline.

PR #56, opened 2022-07-05
remove partner id

Zlaticanin:
# Overview
Removing the partner ID from user-agent string in Azure auth plugin because Microsoft is unable to track these ID's. This change doesn't affect user experience at all.

### DIFF
--- a/util.go
+++ b/util.go
@@ -2,12 +2,10 @@ package azureauth
 
 import (
 	"encoding/json"
-	"fmt"
 	"strings"
 	"time"
 
 	"github.com/hashicorp/vault/sdk/helper/useragent"
-	"github.com/hashicorp/vault/sdk/version"
 )
 
 // Using the same time parsing logic from https://github.com/coreos/go-oidc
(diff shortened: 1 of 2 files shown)


util.go
```diff
@@ -2,12 +2,12 @@ package azureauth
 
 import (
 	"encoding/json"
-	"fmt"
 	"strings"
 	"time"
 
 	"github.com/hashicorp/vault/sdk/helper/useragent"
-	"github.com/hashicorp/vault/sdk/version"
+
```
austingebauer, line 9: nit: same here I think you can remove this blank line

util.go
```diff
@@ -45,24 +45,13 @@ func strListContains(haystack []string, needle string) bool {
...
-func userAgent() string {
-	ua := useragent.String()
-
-	// ent has many version variations, so if it's not "dev" or "" we'll assume
-	// it's an enterprise variation
-	guid := ossVaultGUID
-	ver := version.GetVersion()
-	if ver.VersionMetadata != "" && ver.VersionMetadata != "dev" {
-		guid = entVaultGUID
+func userAgent(pluginEnv *logical.PluginEnvironment) string {
```
austingebauer, line 51: nit: Can update the comment above. Can take it from the old PR I had:
```
// userAgent returns the User-Agent header that's included in HTTP requests
// made by this plugin. Returns an empty string if the pluginEnv is nil.
```

austingebauer (approved):
Just a few small comments. LGTM otherwise 👍

PR #58, opened 2022-07-22
Add release-metadata.hcl for CRT

swenson:


.release/release-metadata.hcl
```diff
@@ -0,0 +1,6 @@
+url_docker_registry_dockerhub = "https://hub.docker.com/r/hashicorp/vault-plugin-auth-azure"
+url_docker_registry_ecr       = "https://gallery.ecr.aws/hashicorp/vault-plugin-auth-azure"
```
tvoran, line 2: I think we can omit these since there's no image being built. At least that seems to be ok in VLE: https://github.com/hashicorp/vault-lambda-extension/blob/main/.release/release-metadata.hcl

swenson: :+1:

swenson:
Okay, I *think* this should work when it is merged. 🤞

Makefile
```diff
@@ -16,7 +17,7 @@ default: dev
 quickdev: generate
 	@CGO_ENABLED=0 go build -i -tags='$(BUILD_TAGS)' -o bin/${TOOL}
 dev: generate
-	@CGO_ENABLED=0 BUILD_TAGS='$(BUILD_TAGS)' VAULT_DEV_BUILD=1 sh -c "'$(CURDIR)/scripts/build.sh'"
+	@CGO_ENABLED=0 BUILD_TAGS='$(BUILD_TAGS)' VERSION='$(VERSION)' VAULT_DEV_BUILD=1 sh -c "'$(CURDIR)/scripts/build.sh'"
```
tvoran, line 20: Should the version tag be added to the `bin` target as well?

Makefile
```diff
@@ -51,4 +52,7 @@ deps:
 	@echo "==> Updating deps for ${TOOL}"
 	@dep ensure -update
 
-.PHONY: bin default generate test bootstrap fmt deps
+version:
+	@echo $(VERSION)
```
tvoran, line 56: I don't see this being used in `.github/workflows/build.yaml` yet, so is this and the `version/version.go` additions just for the future?

Makefile
swenson: Good point, I'll add it.

Makefile
swenson: It's also useful for debugging to see what the detected value of `VERSION` is.

tvoran: Yeah, it seems we've been determining the release version a couple different ways when converting to CRT. vault-k8s and vault-lambda-extension get the version as [input when calling the github action to build](https://github.com/hashicorp/vault-k8s/blob/main/.github/workflows/build.yml#L36-L41), and just set it at compile time in [version/version.go](https://github.com/hashicorp/vault-k8s/blob/2409906e50020532a330e98280e7da9217a2dc70/version/version.go#L8), so that we don't have to update version.go every time we release. vault-ssh-helper is using the [`make version` style](https://github.com/hashicorp/vault-ssh-helper/blob/418e742a3a4cd043ea6ee73a78983cc820e46b4d/.github/workflows/build.yml#L28-L32), like you've started here.

It seems a little redundant to me to have the version tagged on github and statically set in version.go, but maybe that's a better way to go for plugins that have release branches (vault-k8s and VLE do not). CC'ing @tomhjp for his thoughts too.

tomhjp: You're right that CRT recommends a static version in the repo. However, that introduces a fair bit of toil when multiplied by the number of plugin repos, so my 2c would be to dynamically set the version when we trigger releases. Picking the correct version at release time is also toil when you need to do it for N repos and probably multiple release branches, but at least that bit should be pretty trivial and safe to automate.

Perhaps this is worth getting together to hash out the whole design though, because it's a non-trivial system and we haven't really planned it holistically yet.

swenson: makes me want to switch to a monorepo. Or at least have a release monorepo just git clones all the plugins, builds releases, and pushes tags back?

tomhjp: Yeah I see a release monorepo in our future.

swenson:
Closing since we will not be using CRT

PR #61, opened 2022-09-27
Allow finer grain resource verification for path login

k-simons:
# Overview
- Currently resource verification done in the `verifyResource` function of `path_login.go` does not allow for finer grain validation if the registered vault role only contains a subset of the bound resource types
- We have use cases in which we bind either set `BoundSubscriptionsIDs` or `BoundResourceGroups`, however when we are logging into use that role, we do not expect the login to come from a stable/singular `VM` or `VSS`
- Because we are not registering the VaultRole with a `BoundScaleSets` or a `BoundLocation`, we would expect to be able to login just by providing the given `subscriptionsID` or the `resourceGroups` that the role was registered with
- Currently we are not able to do that, because the `verifyResource` function asserts that a `vmssName` or `vmName` must be present even if there is not a underlying bound VM in the configured vault role
- `verifyResource` side steps all validation in the case of:
```
	if len(role.BoundResourceGroups) == 0 && len(role.BoundSubscriptionsIDs) == 0 && len(role.BoundLocations) == 0 && len(role.BoundScaleSets) == 0 {
		return nil
}
```
, however we would still like to configure our registered vault role with a bound subscriptionId or bound resourceID
- It is not currently possible to do so
- This PR makes the resource verification more fine grain, we will do and require the resource validation if it is configured in the configured vault role, however if it is not, that resource verification will not be ran
- End user should expect all current log in flows to work as expected. Additionally, new log in flows should also pass resource verification if the underlying vault role is not bound to a location or a scale set and the user does not provide one

# Design of Change

[ ] [Issue #1234](https://github.com/hashicorp/vault/issues/1234)
[ ] [PR #1234](https://github.com/hashicorp/vault/pr/1234)

# Contributor Checklist
[ ] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
[My Docs PR Link](link)
[Example](https://github.com/hashicorp/vault/commit/2715f5cec982aabc7b7a6ae878c547f6f475bba6)
[ ] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
[ ] Backwards compatible

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -230,16 +230,29 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
 	return nil
 }
 
-func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, resourceGroupName, vmName string, vmssName string, claims *additionalClaims, role *azureRole) error {
-	// If not checking anything with the resource id, exit early
-	if len(role.BoundResourceGroups) == 0 && len(role.BoundSubscriptionsIDs) == 0 && len(role.BoundLocations) == 0 && len(role.BoundScaleSets) == 0 {
-		return nil
-	}
+type verifyFuncAlias func(ctx context.Context, subscriptionID, resourceGroupName, vmName string, vmssName string, claims *additionalClaims, role *azureRole) error
 
-	if subscriptionID == "" || resourceGroupName == "" {
-		return errors.New("subscription_id and resource_group_name are required")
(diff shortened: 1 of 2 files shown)


k-simons:
Tagging @tomhjp as the last person who merged something in this repo. Apologizes in advance if that isn't correct! I could not find a contribution guide/something similar for what the review policy is

okushchenko:
@austingebauer any chance you might take a look at this PR? This would be very helpful for my use-case with Azure Auth Backend.

Issue #62, opened 2022-09-30
Support VMSS Flex Authentications


dkiser:
**Desired Behavior**
As a `vault-plugin-auth-azure` user, I Should Be Able to provide instance identity credentials from a VMSS **Flex** Virtual Machine to the [login endpoint](https://www.vaultproject.io/api-docs/auth/azure#login) So That I Can access Vault resources from VMSS **Flex** VMs in addition to VMSS VMs.

**Error Shown**
`token object id does not match virtual machine identities` from [here](https://github.com/hashicorp/vault-plugin-auth-azure/blob/0e097203989e2bbc4e86cb6034808e8b42e32190/path_login.go#L314)

**Steps to reproduce**
1. Install/configure the `vault-pluign-auth-azure` plugin
2. Create a VMSS *Flex* definition in Azure and launch VMs from this definition
3. Create a role for a VMSS to mach the VMSS name/attributes from step 2.
4. From a VMSS Flex VM, post the instance metadata identity JWT token to the [login API](https://www.vaultproject.io/api-docs/auth/azure#login) with the appropriate role.
5. View the error response above

dkiser:
From initial investigations, the [VitualMachineScaleSetIdentityUserAssignedIdentitiesValue](https://github.com/Azure/azure-sdk-for-go/blob/c4b71e1147922015740c80018f57fd6c2cd41ebd/services/compute/mgmt/2021-11-01/compute/models.go#L19228) utilized in `vault-plugin-auth-azure` [here](https://github.com/hashicorp/vault-plugin-auth-azure/blob/0e097203989e2bbc4e86cb6034808e8b42e32190/path_login.go#L276) is never populated from the Azure API response to `GET` for a VMSS object when that VMSS is defined as **Flex**. 

This appears to result in the for loop [here](https://github.com/hashicorp/vault-plugin-auth-azure/blob/0e097203989e2bbc4e86cb6034808e8b42e32190/path_login.go#L276) never populating the `principalIDs` map correctly to be validated [here](https://github.com/hashicorp/vault-plugin-auth-azure/blob/0e097203989e2bbc4e86cb6034808e8b42e32190/path_login.go#L314), resulting in the thrown error.

Example Object responses from the Azure Compute API for VMSS with a User Assigned Identity, and **NO Flex** enabled
```
{
    "name": "**SCRUBBED**",
[...]

PR #63, opened 2022-10-03
Support VMSS Flex Authentications

zmyzheng:
# Overview
Support VMSS Flex Authentications

# Design of Change
When using VMSS Flex with user assigned identity, the service principle Id cannot be dirrectly retrieved from Get-VMSS call. In this PR, we get the user assigned managed identity id and use MSI client to retrieve the service principle info. This change can be applied to both VMSS Uniform and VMSS Flex

 Fix [Issue #62 ](https://github.com/hashicorp/vault-plugin-auth-azure/issues/62)

# Contributor Checklist
[-] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
[My Docs PR Link](link)
[Example](https://github.com/hashicorp/vault/commit/2715f5cec982aabc7b7a6ae878c547f6f475bba6)
[-] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
[-] Backwards compatible

### DIFF
--- a/azure.go
+++ b/azure.go
@@ -11,8 +11,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/Azure/azure-sdk-for-go/profiles/latest/msi/mgmt/msi"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/msi/mgmt/2018-11-30/msi"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
--- a/azure.go
+++ b/azure.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/msi/mgmt/msi"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
@@ -32,6 +33,10 @@ type vmssClient interface {
 	Get(ctx context.Context, resourceGroup, vmssName string, expandTypes compute.ExpandTypesForGetVMScaleSets) (compute.VirtualMachineScaleSet, error)
 }
 
+type msiClient interface {
+	Get(ctx context.Context, resourceGroup, resourceName string) (result msi.Identity, err error)
+}
+
 type tokenVerifier interface {
 	Verify(ctx context.Context, token string) (*oidc.IDToken, error)
 }
@@ -40,6 +45,7 @@ type provider interface {
 	Verifier() tokenVerifier
 	ComputeClient(subscriptionID string) (computeClient, error)
 	VMSSClient(subscriptionID string) (vmssClient, error)
+	MSIClient(subscriptionID string) (msiClient, error)
 }
 
 type azureProvider struct {
@@ -133,6 +139,19 @@ func (p *azureProvider) VMSSClient(subscriptionID string) (vmssClient, error) {
 	}
 
 	client := compute.NewVirtualMachineScaleSetsClientWithBaseURI(p.settings.Environment.ResourceManagerEndpoint, subscriptionID)
+	client.Authorizer = authorizer
+	client.Sender = p.httpClient
+	client.AddToUserAgent(userAgent(p.settings.PluginEnv))
+	return client, nil
+}
+
+func (p *azureProvider) MSIClient(subscriptionID string) (msiClient, error) {
+	authorizer, err := p.getAuthorizer()
+	if err != nil {
+		return nil, err
+	}
+
+	client := msi.NewUserAssignedIdentitiesClientWithBaseURI(p.settings.Environment.ResourceManagerEndpoint, subscriptionID)
 	client.Authorizer = authorizer
 	client.Sender = p.httpClient
 	client.AddToUserAgent(userAgent(p.settings.PluginEnv))
--- a/azure_test.go
+++ b/azure_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/msi/mgmt/msi"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"github.com/coreos/go-oidc"
 )
@@ -44,6 +45,10 @@ type mockVMSSClient struct {
 	vmssClientFunc func(vmssName string) (compute.VirtualMachineScaleSet, error)
 }
 
+type mockMSIClient struct {
+	msiClientFunc func(resourceName string) (msi.Identity, error)
+}
+
 func (c *mockComputeClient) Get(_ context.Context, _, vmName string, _ compute.InstanceViewTypes) (compute.VirtualMachine, error) {
 	if c.computeClientFunc != nil {
 		return c.computeClientFunc(vmName)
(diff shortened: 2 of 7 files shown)


path_login.go
```diff
@@ -273,8 +274,24 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 			principalIDs[to.String(vmss.Identity.PrincipalID)] = struct{}{}
 		}
 		// if not, look for user-assigned identities
-		for _, userIdentity := range vmss.Identity.UserAssignedIdentities {
-			principalIDs[to.String(userIdentity.PrincipalID)] = struct{}{}
+		for userIdentityID := range vmss.Identity.UserAssignedIdentities {
```
austingebauer, line 277: Do you happen to know if it's intentional that the principal ID is omitted for only VMSS Flex? It seems a bit strange that it is missing for Flex but is included for Uniform Orchestration Mode. I'm curious why that might be the case.

austingebauer (commented):
Thanks for opening this PR, @zmyzheng. I've left some initial feedback and questions.

path_login.go
```diff
@@ -273,8 +274,24 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
+			var elements []string = strings.Split(userIdentityID, "/")
+			msiSubscriptionID := elements[2]
+			msiResourceGroupName := elements[4]
+			msiResourceName := elements[8]
```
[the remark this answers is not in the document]
zmyzheng: Sounds good, I added a lengh check before indexing

path_login.go
```diff
@@ -273,8 +274,24 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
+
+			msiClient, err := b.provider.MSIClient(msiSubscriptionID)
+			if err != nil {
+				return errwrap.Wrapf("unable to create msi client: {{err}}", err)
```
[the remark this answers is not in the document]
zmyzheng: Done

path_login.go
```diff
@@ -273,8 +274,24 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
+			}
+			userIdentity, err := msiClient.Get(ctx, msiResourceGroupName, msiResourceName)
+			if err != nil {
+				return errwrap.Wrapf("unable to retrieve user assigned identity metadata: {{err}}", err)
```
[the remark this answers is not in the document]
zmyzheng: Done

path_login.go
```diff
@@ -273,8 +274,24 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 			principalIDs[to.String(vmss.Identity.PrincipalID)] = struct{}{}
 		}
 		// if not, look for user-assigned identities
-		for _, userIdentity := range vmss.Identity.UserAssignedIdentities {
-			principalIDs[to.String(userIdentity.PrincipalID)] = struct{}{}
+		for userIdentityID := range vmss.Identity.UserAssignedIdentities {
+			var elements []string = strings.Split(userIdentityID, "/")
```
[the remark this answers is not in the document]
zmyzheng: Done

azure.go
```diff
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/msi/mgmt/msi"
```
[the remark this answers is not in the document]
zmyzheng: Thanks for raising this issue. My understanding is packages under /profiles are auto-generated and refer to the packages under /services. I just chhanged reference here to directly use packages from "github.com/Azure/azure-sdk-for-go/services/msi/mgmt/2018-11-30/msi"

path_login.go
zmyzheng: Yes, it's intentional that the principal ID is omitted for VMSS Flex. This is because this information cannot be directly retrieved from VMSS DB. The Uniform API is designed previously, in which the VMSS needs to query the identity API behind the sense. Now for Flex, to keep the VMSS orchestrator as light, and avoid dependency on another resource, the VMSS engineer is enforcing the API not to query identity any more. But for Uniform, we still do that just for backward compatibility.

austingebauer: Thanks for the explanation, @zmyzheng. Makes sense 👍

path_login.go
```diff
@@ -273,8 +274,27 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 			principalIDs[to.String(vmss.Identity.PrincipalID)] = struct{}{}
 		}
 		// if not, look for user-assigned identities
-		for _, userIdentity := range vmss.Identity.UserAssignedIdentities {
-			principalIDs[to.String(userIdentity.PrincipalID)] = struct{}{}
+		for userIdentityID := range vmss.Identity.UserAssignedIdentities {
```
austingebauer, line 277: I'm wondering if we should check if the user-assigned identity has a non-nil principal ID first and fall back to looking it up using the MSI client. This way our current users relying on VMSS uniform won't be broken by needing to update the secrets engine credentials with an additional role assignment. The new MSI API call requires `Microsoft.ManagedIdentity/userAssignedIdentities/read`. We don't want to break our current VMSS uniform users by introducing the code that works for flex.

For example, we could do something like:
```go
// if not, look for user-assigned identities
for userIdentityID, userIdentity := range vmss.Identity.UserAssignedIdentities {
	// Principal ID is not nil for VMSS uniform orchestration mode
	if userIdentity.PrincipalID != nil {
		principalIDs[to.String(userIdentity.PrincipalID)] = struct{}{}
		continue
	}

	// Principal ID is nil for VMSS uniform orchestration mode, so we 
	// must look up the user-assigned identity using the MSI client
	elements := strings.Split(userIdentityID, "/")
	if len(elements) < 9 {}
	...
}
```

path_login.go
```diff
@@ -273,8 +274,27 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
+			elements := strings.Split(userIdentityID, "/")
+			if len(elements) < 9 {
+				return fmt.Errorf("ubable to parse the userIdentityID: %s", userIdentityID)
```
austingebauer, line 280:
```suggestion
				return fmt.Errorf("unable to parse the userIdentityID: %s", userIdentityID)
```

path_login.go
zmyzheng: Agree, that's good point! I just changed the code based on your suggestion.

path_login.go
zmyzheng: Good catch! Thanks

path_login.go
```diff
@@ -273,8 +274,33 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
+			}
+
+			elements := strings.Split(userIdentityID, "/")
+			if len(elements) < 9 {
+				return fmt.Errorf("unable to parse the userIdentityID: %s", userIdentityID)
+			}
+			msiSubscriptionID := elements[2]
+			msiResourceGroupName := elements[4]
+			msiResourceName := elements[8]
+
```
austingebauer, line 291:
```suggestion
	// Principal ID is nil for VMSS flex orchestration mode, so we 
	// must look up the user-assigned identity using the MSI client
```

austingebauer: Suggest to add this comment to explain the non-obvious bits of why we're doing this for future readers.

zmyzheng: Sure

azure.go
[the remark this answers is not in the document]
austingebauer: Got it. Thanks! Looks like `2018-11-30` was the latest for this from the import in profiles.

azure_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/msi/mgmt/msi"
```
austingebauer, line 10: We should probably change this to use `/services` as we did in other places.

austingebauer: Might want to run `go mod tidy` after that as well.

path_login.go
```diff
@@ -273,8 +274,35 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
+		for userIdentityID, userIdentity := range vmss.Identity.UserAssignedIdentities {
+			// Principal ID is not nil for VMSS uniform orchestration mode
+			if userIdentity.PrincipalID != nil {
+				principalIDs[to.String(userIdentity.PrincipalID)] = struct{}{}
+				continue
+			}
+
+			elements := strings.Split(userIdentityID, "/")
+			if len(elements) < 9 {
+				return fmt.Errorf("unable to parse the userIdentityID: %s", userIdentityID)
```
austingebauer, line 286: Suggest to not use the variable name in the error for users:

```suggestion
				return fmt.Errorf("unable to parse the user-assigned identity resource ID: %s", userIdentityID)
```

azure_test.go
zmyzheng: Done

path_login.go
zmyzheng: Done

austingebauer (approved):
LGTM. Thanks again for this contribution, @zmyzheng.

I'm happy to contribute the [Vault documentation](https://developer.hashicorp.com/vault/docs/auth/azure) for this if you haven't started on it already. Let me know.

zmyzheng:
> LGTM. Thanks again for this contribution, @zmyzheng.
> 
> I'm happy to contribute the [Vault documentation](https://developer.hashicorp.com/vault/docs/auth/azure) for this if you haven't started on it already. Let me know.

Thanks @austingebauer so much for reviewing the PR. It would be great if you can update the Vault document.

austingebauer:
@zmyzheng - Happy to help. I will update the Vault documentation 👍

PR #65, opened 2022-11-01
Upgrades dependencies, adds environment bootstrap

austingebauer:
# Overview

This PR upgrades dependencies, tooling, and adds development environment automation. Changes in this PR align this repo with tooling in the Vault plugin scaffold.

Local environment bootstrap usage:
```sh
make setup-env
source ./bootstrap/terraform/local_environment_setup.sh
make configure
make teardown-env
```

Test Results:
```sh
$ TESTARGS='-count=1' make test
==> Checking that code complies with gofumpt requirements...
CGO_ENABLED=0 go test ./... -count=1 -timeout=20m
ok      github.com/hashicorp/vault-plugin-auth-azure    0.240s
?       github.com/hashicorp/vault-plugin-auth-azure/cmd/vault-plugin-auth-azure        [no test files]
```

Note that there aren't acceptance tests that actually use Azure resources in this plugin, so that's omitted from this description.

# Contributor Checklist
- [x] Backwards compatible


austingebauer:
I still need to update the readme. Will do that tomorrow.

bootstrap/configure.sh
```diff
@@ -0,0 +1,43 @@
...
+    bound_subscription_ids="${SUBSCRIPTION_ID}" \
+    bound_resource_groups="${RESOURCE_GROUP_NAME}"
+    
+# Login using the access token
+vault write auth/"${PLUGIN_PATH}"/login \
+    role="dev-role" \
+    jwt="${ACCESS_TOKEN_JWT}" \
+    subscription_id="${SUBSCRIPTION_ID}" \
+    resource_group_name="${RESOURCE_GROUP_NAME}" \
+    vm_name="${VM_NAME}"
```
austingebauer, line 43: I could probably comment this out but leave it as an example. I _think_ it'll be helpful to see a full working example here. Let me know!

bootstrap/terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,99 @@
+# This file is maintained automatically by "terraform init".
```
austingebauer, line 1: Intentionally checked in based on guidance from Terraform docs.

bootstrap/terraform/vm.tf
```diff
@@ -0,0 +1,218 @@
...
+  required_providers {
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "3.29.1"
+    }
+    azuread = {
+      source  = "hashicorp/azuread"
+      version = "2.30.0"
+    }
+  }
```
austingebauer, line 11: Open to suggestions on pinning these to specific version or not doing so. Let me know!

bootstrap/terraform/scripts/imds_access_token.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+# External data source script for obtaining an access token JWT from the Azure 
+# Instance Metadata Service (IMDS) endpoint. The access token is used as the form
+# of authentication for the Vault Azure Auth Method. For details, see
+# - https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http
+#- https://registry.terraform.io/providers/hashicorp/external/latest/docs/data-sources/data_source
+# - https://developer.hashicorp.com/vault/api-docs/auth/azure#jwt
```
austingebauer, line 8: This seems to work well, but there could be a better way to do this. I'm not super Terraform savvy. Open to ideas!

bootstrap/terraform/vm.tf
```diff
@@ -0,0 +1,218 @@
...
+
+  security_rule {
+    name                       = "ssh"
+    priority                   = 100
+    direction                  = "Inbound"
+    access                     = "Allow"
+    protocol                   = "Tcp"
+    source_port_range          = "*"
+    destination_port_range     = "22"
+    source_address_prefix      = chomp(data.http.my_ip.body)
```
fairclothjm, line 115: Getting a deprecation notice from terraform when running `make setup-env` for the body attribute. Maybe we should use [`response_body`](https://registry.terraform.io/providers/hashicorp/http/latest/docs/data-sources/http#body) instead?

bootstrap/terraform/vm.tf
```diff
@@ -0,0 +1,218 @@
...
+      version = "3.29.1"
+    }
+    azuread = {
+      source  = "hashicorp/azuread"
+      version = "2.30.0"
+    }
+  }
+}
+
+variable "ssh_public_key_path" {
```
fairclothjm, line 14: Maybe this needs to be all caps? Initially, I set `export TF_VAR_SSH_PUBLIC_KEY_PATH=~/.ssh/id_ed25519.pub` but it still used the default. So I did `export TF_VAR_ssh_public_key_path=~/.ssh/id_ed25519.pub` and that picked it up.

Incidentally, Only RSA SSH keys are supported by Azure. TIL

README.md
```diff
@@ -62,59 +62,115 @@ This will put the plugin binary in the `bin` and `$GOPATH/bin` folders. `dev`
...
 
-Successfully enabled 'plugin' at 'azure'!
+Successfully enabled 'plugin' at 'azure-plugin'!
+```
+
+### Azure Environment Setup
+
+A Terraform [configuration](bootstrap/terraform) is included in this repository that
+automates provisioning of Azure resources necessary to configure and authenticate
+using the auth method.
```
fairclothjm, line 120: nit: Might be good to include a list of resources that this config will create? Something like a bulleted list similar to the VLE: https://github.com/hashicorp/vault-lambda-extension/tree/main/quick-start#vault-lambda-extension-quick-start

fairclothjm:
After fixing my ssh key env this terraform setup and tests all worked great!

bootstrap/configure.sh
fairclothjm: I don't think it hurts to leave this here. It is nice to have the full example

bootstrap/terraform/vm.tf
fairclothjm: based on what I am reading [here](https://developer.hashicorp.com/terraform/tutorials/configuration-language/provider-versioning) +1 to pinning to a version

fairclothjm: It would be nice if we could automate bumping a pinned version to run in CI. Maybe that is a thing?

bootstrap/terraform/vm.tf
austingebauer: Whoops, that was my mistake in the readme. Fixed in [690e71b](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/690e71bbea0a74edb7200e6a38562f0edf7071c6).

Interesting, TIL as well!

bootstrap/terraform/vm.tf
austingebauer: Good find. Done in [56208c3](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/56208c3569d81279d1c75247f8ed8d4c118741f0).

README.md
austingebauer: Good idea! I'll include something similar to that 👍

austingebauer: Added in [52970b2](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/52970b2117fb589bb36c0959d27ad9b03bb3855a). Thanks!

fairclothjm (approved):
👍

bootstrap/terraform/vm.tf
```diff
@@ -0,0 +1,216 @@
...
+    azuread = {
+      source  = "hashicorp/azuread"
+      version = "2.30.0"
+    }
+    http = {
+      source  = "hashicorp/http"
+      version = "3.2.0"
+    }
+  }
+}
```
maxcoulombe, line 16: Should we specify a minimum version instead of a specific version?
https://developer.hashicorp.com/terraform/language/providers/requirements#version-constraints
`version = ">= 3.29.1"` etc.

Since this is for local development and possibly CI steps, and not to upgrade production environments, I'd be fine with always using the latest versions personally.

bootstrap/terraform/vm.tf
```diff
@@ -0,0 +1,216 @@
...
+
+data "azurerm_client_config" "current" {}
+provider "azuread" {}
+provider "azurerm" {
+  features {}
+}
+
+resource "azuread_application" "vault_azure_app" {
+  display_name = "vault_azure_app"
+}
```
maxcoulombe, line 26: Not a big deal as it is a small local setup config but if you ever work on larger production Terraform projects it is recommended to [not repeat the resource type in it's name](https://www.terraform-best-practices.com/naming). Like "...application" & "...app" & other examples below.

It is also recommended by HashiCorp to [prefix the resource name with the provider's name](https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/naming).

And a personal preference of mine is to also have a "namespace" prefix to group resource that depend on one another. And name the resource based on it's purpose in the overall project.

Given all this resource names could be:
`vault_azure_tests`
With an extra section if you need disambiguation between resources of the same type, for example with assignments:
```
vault_azure_tests_system
vault_azure_tests_user
```

But yeah naming is always a topic with tons of grey area so I don't have a strong opinion, just some pointers from docs.

bootstrap/terraform/vm.tf
```diff
@@ -0,0 +1,216 @@
...
+  scope                = azurerm_linux_virtual_machine.vault_azure_vm.id
+  principal_id         = azurerm_linux_virtual_machine.vault_azure_vm.identity[0].principal_id
+}
+
+# Uncomment below for user assigned managed identity
+# resource "azurerm_user_assigned_identity" "vault_azure_uid" {
+#   name                = "vault_azure_uid"
+#   resource_group_name = azurerm_resource_group.vault_azure_rg.name
+#   location            = azurerm_resource_group.vault_azure_rg.location
+# }
```
maxcoulombe, line 48: Could/should we create resources for both identity types at the same time so contributors do not need to modify the Terraform project based on which method they want to setup?

tools/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/hashicorp/vault-plugin-scaffolding/tools
+
+go 1.17
```
maxcoulombe, line 3: 1.19 here as well?

maxcoulombe (commented):
Waiting on helpdesk to get access to an Azure account to test the full setup but it looks good!

CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+
+* Adds support for Virtual Machine Scale Set Flex Authentication [GH-63](https://github.com/hashicorp/vault-plugin-auth-azure/pull/63)
+
+IMPROVEMENTS:
+
+* Updates dependencies: `github.com/Azure/azure-sdk-for-go v67.0.0+incompatible`, 
+  `github.com/Azure/go-autorest/autorest v0.11.28`, `github.com/hashicorp/go-hclog v1.3.1`,
+  `github.com/hashicorp/vault/api v1.8.2`, `github.com/hashicorp/vault/api v1.8.2`, 
+  `github.com/hashicorp/vault/sdk v0.6.1`, `golang.org/x/oauth2 v0.1.0` 
+  [[GH-65](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65)]
```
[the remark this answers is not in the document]
austingebauer: Done in [1439ffb](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/1439ffb14212ff47143bfe0a97e073edf797a84a) 👍

README.md
```diff
@@ -62,59 +62,130 @@ This will put the plugin binary in the `bin` and `$GOPATH/bin` folders. `dev`
...
+$ source ./bootstrap/terraform/local_environment_setup.sh
+```
+
+Next, run the `make configure` target to register, enable, and configure the plugin with
+your local Vault instance. You can specify the plugin name, plugin directory, and mount
+path. Default values from the Makefile will be used if arguments aren't provided.
+
+```sh
+$ PLUGIN_NAME=vault-plugin-auth-azure \
+  PLUGIN_DIR=path/to/plugin/directory \
```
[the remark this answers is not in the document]
austingebauer: Changed in [3bded77](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/3bded77619773ff102c4816a602733ad547279b0).

bootstrap/terraform/vm.tf
austingebauer: Ya that's a good point. It'd be nice if these could just use whatever the latest version is so that we don't need to worry about upgrading them over time. I'll use the `>=` 👍

austingebauer: Done in [3d2cbee](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/3d2cbeee231ccba733f24eefb79a8d12c73cb822).

bootstrap/terraform/vm.tf
austingebauer: Some convo around this at https://github.com/hashicorp/vault-plugin-auth-azure/pull/65#discussion_r1012302796. I think it's okay for TF to automatically download the latest since this is just for testing / development environments.

tools/go.mod
austingebauer: Good catch. Done in [985f9d4](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/985f9d49276d7e45f3b267ce282ec10a497d0acc).

bootstrap/terraform/vm.tf
maxcoulombe: If you omit the required_providers block entirely terraform init will use the latest version at all times.

It's a trade-off between reproduceability/stability & toil of managing versions. 

I think for customer facing infrastructure changes having set version is a no brainer for safety. But for QoL-projects where a failed run is not a big deal I'm not against removing this block completely so we catch our own regressions.

maxcoulombe:
LGTM, important comments are addressed and as for:
* TF resource naming
* leaving the required_provider block
* provisioning resources for both identify types at all time 
These are more questions/observations than suggestions. Happy to chat if you have questions.

bootstrap/terraform/vm.tf
fairclothjm: +1 @maxcoulombe good points

bootstrap/terraform/vm.tf
austingebauer: Yep, looks like we can create both (system-assigned and user-assigned) and let the user choose when getting the access token from IMDS. Done in [f2eb6a5](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/f2eb6a53c924fdfce8e140e9e00c8c93e121cee1).

bootstrap/terraform/vm.tf
austingebauer: Nice! Thanks for this info. I was unsure of what to do with these names. I'll change them 👍

bootstrap/terraform/vm.tf
austingebauer: +1 - I'll remove the required_provides block entirely. This will be consistent with what's in GCP secrets/auth too.

austingebauer: Done in [cd179cd](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/cd179cdc859e7e7aa5dc124afe038562930399f9).

bootstrap/terraform/vm.tf
austingebauer: Done in [c3e62ec](https://github.com/hashicorp/vault-plugin-auth-azure/pull/65/commits/c3e62ec7d57e031fa73f5f262576bb55d5d45c99).

Issue #17, opened 2019-01-07
Support for AppService/Azure Functions


lenisha:
Hi,

Could MSI for running in AppService be supported? Currently it's only VM or VMSS:

curl \
    --request POST \
    --data '{"role": "dev-role", "jwt": "<msi token>'", "subscription_id":"<id of subscription>","resource_group_name":"node-vault", "**vm_name**":"vault"}' \
    http://127.0.0.1:8200/v1/auth/azure/login

chrishoffman:
Thanks for the info.  I assume we would pin it to a list of `appid`s?  I think this should be possible and probably pretty easy to add if anyone wants to take a shot at a PR before we get to this.

softveda:
I think this is very critical as many cloud native teams are using Azure Functions, App Services and AKS to deploy applications that integrate with azure resources using managed identity. Apart from System assigned managed identities, User assigned identities must also be supported.

mwieczorek:
cc @calvn, @tvoran, @jasonodonnell, @kalafut, @benashz 
(I'm adding you here because you were authors/reviewers of recent PRs in this repo. Maybe you can confirm if you have 'bandwidth' to review such PR?)

mwieczorek:
I looked at the code and would like to confirm how this should be implemented.

### Like for VM/VMSS

1. Add new optional parameter in [login api](https://www.vaultproject.io/api/auth/azure#login), f.e. `function_name`
2. Add new optional parameter for [role](https://www.vaultproject.io/api/auth/azure#create-role), f.e. `bound_function`
3. In code, add new `case` [in this switch statement](https://github.com/hashicorp/vault-plugin-auth-azure/blob/647836b8118adba7414d4d409059586f84f1f270/path_login.go#L226) for handling Azure Functions

### Simplify

Allow [login operation](https://www.vaultproject.io/api/auth/azure#login) when:
a) `vm_name` / `vmss_name` **are not passed** (now one of them is required)
b) `bound_service_principal_ids` **is set** for [selected role](https://www.vaultproject.io/api/auth/azure#bound_service_principal_ids)
c) `oid` claim from JWT **is one of** `bound_service_principal_ids`

In this case validation of: `bound_service_principal_ids`, `bound_group_ids`, `bound_subscription_ids`, `bound_resource_groups` can be done based on claims from the token, and no Azure API calls are required.
In case the role has `bound_scaleset` - login will fail
In case the role has `bound_location` - login could fail too

---

[...]

PR #71, opened 2022-12-02
extend managed identities support

fairclothjm:
# Overview
Currently, the Vault Azure auth method can only authenticate Azure VMs
or VM Scale Sets (VMSS) using Azure’s managed identity feature. This PR
implements a more general solution that should allow any Azure resource
that supports managed identities to authenticate with Vault.

The main change here is that we will require the `resource_id` on login instead of `vm_name` or `vmss_name`:

```
vault write auth/vault-plugin-auth-azure/login \
    role="dev-role" \
    jwt="${ACCESS_TOKEN_JWT}" \
    subscription_id="${SUBSCRIPTION_ID}" \
    resource_group_name="${RESOURCE_GROUP_NAME}" \
    resource_id="${RESOURCE_ID}"
```

resolves https://github.com/hashicorp/vault-plugin-auth-azure/issues/17

Related docs PR: https://github.com/hashicorp/vault/pull/18357

# Contributor Checklist
- [x] Backwards compatible


azure.go
```diff
@@ -190,7 +180,32 @@ func (p *azureProvider) MSIClient(subscriptionID string) (msiClient, error) {
...
+	clientOptions := p.getClientOptions()
+	client, err := armresources.NewClient(subscriptionID, cred, clientOptions)
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+func (p *azureProvider) getClientOptions() *arm.ClientOptions {
```
austingebauer, line 207: 👍

bootstrap/terraform/function/function-project/vlt-auth-func-test/__init__.py
```diff
@@ -0,0 +1,62 @@
+import logging
+import requests
+import os
+
+import azure.functions as func
+
+
+def main(req: func.HttpRequest) -> func.HttpResponse:
+    # return get_sys_assigned_id_token()
+    return get_user_assigned_id_token()
```
austingebauer, line 10: Nice! Looks like this works for both system and user assigned.

bootstrap/terraform/function/main.tf
```diff
@@ -0,0 +1,154 @@
...
+  features {
+    resource_group {
+      # the function is published via the Azure CLI so we will ensure any
+      # new resources are also clean up on destroy
+      prevent_deletion_if_contains_resources = false
+    }
+  }
+}
+
+resource "random_pet" "main" {
```
austingebauer, line 13: TIL random_pet

bootstrap/terraform/vm.tf
```diff
@@ -157,6 +157,7 @@ resource "local_file" "setup_environment_file" {
 export ACCESS_TOKEN_JWT=${data.external.access_token_jwt.result.access_token}
 export VM_NAME=${azurerm_linux_virtual_machine.vault_azure_vm.name}
 export VM_IP_ADDRESS=${azurerm_public_ip.vault_azure_pub_ip.ip_address}
+export RESOURCE_ID=${azurerm_linux_virtual_machine.vault_azure_vm.id}
```
austingebauer, line 160: I'm guessing this resource ID approach works for VM-based managed identity too?

path_login.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/hashicorp/vault/sdk/logical"
 )
 
+var resourceClientAPIVersion = "2022-03-01"
```
austingebauer, line 17: Might be worth dropping a comment on here with a link to where these versions can be found? Thinking if we need to update it in the future.

austingebauer (approved):
Just a few minor questions and suggestions. I like the generic approach. Thanks for adding the environment automation too. LGTM 👍

bootstrap/terraform/vm.tf
fairclothjm: yep!

path_login.go
fairclothjm: Unfortunately this is not documented at all except to say that it is the "API version to use". You can find this in the Azure Portal when looking at a resource JSON View. Interestingly the `sdk/resourcemanager/resources/armresources` module is the only one in the entire azure-sdk-for-go that allows setting this magic value. I am wondering if we could set it as empty string to always use the most recent. I just used this value since it is the most recent available in the Azure Portal resource JSON view.

bootstrap/terraform/function/function-project/vlt-auth-func-test/__init__.py
fairclothjm: yep, I tested with both

path_login.go
fairclothjm: I think this is our best bet as far as documentation goes: https://learn.microsoft.com/en-us/rest/api/resources/resources/get-by-id#uri-parameters

I went ahead and changed it to the value defined there and tested that everything works as expected

bootstrap/terraform/function/README.md
```diff
@@ -0,0 +1,90 @@
+# Azure Function terraform setup
```
maxcoulombe, line 1: gg

bootstrap/terraform/function/main.tf
maxcoulombe: TIL as well, this is neat

maxcoulombe (approved):

path_login.go
austingebauer: Ah, gotcha. Thanks for the info and for adding the link in the comment. Looks good to me.

PR #73, opened 2022-12-12
Adds random suffix to resource group for env setup

austingebauer:
This PR adds a random suffix to the resource group created by the environment setup automation. This helps prevent collisions when the automation is being used on a shared Azure subscription. The resource group name will look something like: `vault_azure_tests_38eb1129`.


vinay-gopalan (approved):
Thanks for including this!

fairclothjm (approved):
👍

PR #75, opened 2022-12-16
query supported API versions for resources

fairclothjm:
# Overview
https://github.com/hashicorp/vault-plugin-auth-azure/pull/71 extended support for managed identities but did not take into account that different resource types will have different API versions that they support. This PR will query the [Providers](https://learn.microsoft.com/en-us/rest/api/resources/providers/get?tabs=HTTP) to determine the supported API versions for the resource type that is being used on login.

# Contributor Checklist
- [x] Backwards compatible

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/vault/sdk/framework"
 	"github.com/hashicorp/vault/sdk/helper/cidrutil"
@@ -15,9 +16,10 @@ import (
 )
 
 // resourceClientAPIVersion is the API version to use for the operation. This
-// is not well documented but we are setting it to the current value defined
-// here: https://learn.microsoft.com/en-us/rest/api/resources/resources/get-by-id#uri-parameters
-var resourceClientAPIVersion = "2021-04-01"
+// is not well documented but supported API version can be queried from the
+// GET Providers endpoint.
+// https://learn.microsoft.com/en-us/rest/api/resources/providers/get?tabs=HTTP
+var resourceClientAPIVersion = "2022-03-01"
 
 func pathLogin(b *azureAuthBackend) *framework.Path {
 	return &framework.Path{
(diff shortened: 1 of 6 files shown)


path_login.go
```diff
@@ -454,3 +462,47 @@ func convertPtrToString(s *string) string {
...
+	if err != nil {
+		return "", fmt.Errorf("unable to create providers client: %w", err)
+	}
+
+	elements := strings.Split(resourceID, "/")
+	if len(elements) < 9 {
+		return "", fmt.Errorf("unable to parse the resource ID: %s", resourceID)
+	}
+	providerNamespace := elements[6]
+	resourceType := elements[7]
```
elsesiy, line 479: Microsoft provides a public API for parsing resource IDs [here](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm#ParseResourceID)

path_login.go
```diff
@@ -454,3 +462,47 @@ func convertPtrToString(s *string) string {
...
+
+	response, err := client.Get(ctx, providerNamespace, nil)
+	if err != nil {
+		return "", fmt.Errorf("unable to get the provider for resource %s: %w", resourceID, err)
```
elsesiy, line 483:
```suggestion
		return "", fmt.Errorf("unable to get the provider for resource %q: %w", resourceID, err)
```

path_login.go
```diff
@@ -454,3 +459,44 @@ func convertPtrToString(s *string) string {
...
+
+// getAPIVersionForResource attempts to query the supported API versions for a
+// given resource.
+func (b *azureAuthBackend) getAPIVersionForResource(ctx context.Context, subscriptionID, resourceID string) (string, error) {
+	client, err := b.provider.ProvidersClient(subscriptionID)
+	if err != nil {
+		return "", fmt.Errorf("unable to create providers client: %w", err)
+	}
+
+	resourceType, err := arm.ParseResourceType(resourceID)
```
fairclothjm, line 471: `resourceType` will have the following structure for an Azure Function:
```
github.com/Azure/azure-sdk-for-go/sdk/azcore/arm.ResourceType {
	Namespace: "Microsoft.Web",
	Type: "sites",
	Types: []string len: 1, cap: 1, ["sites"],
	stringValue: "Microsoft.Web/sites",
}
```

path_login.go
```diff
@@ -297,21 +299,18 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 				continue
 			}
 
-			elements := strings.Split(userIdentityID, "/")
-			if len(elements) < 9 {
-				return fmt.Errorf("unable to parse the user-assigned identity resource ID: %s", userIdentityID)
+			msiID, err := arm.ParseResourceID(userIdentityID)
```
austingebauer, line 302: Nice. TIL `ParseResourceID`.

austingebauer (approved):
LGTM. Thanks @fairclothjm! I'm thinking we'll also want to document if there are any additional read permissions needed for the credential given to Vault for the new providers request being made.

fairclothjm:
@austingebauer Thanks! I will double check the permissions but I don't think there are any additional required permissions.

PR #76, opened 2022-12-20
add caching for resource api version

fairclothjm:
# Overview
#75 adds support to query the [Providers](https://learn.microsoft.com/en-us/rest/api/resources/providers/get?tabs=HTTP) to determine the supported API versions for the resource type that is being used on login.

This PR adds caching for the resource API version so that subsequent logins will not make the same API call more than once.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -23,6 +23,11 @@ type azureAuthBackend struct {
 	l sync.RWMutex
 
 	provider provider
+
+	// resourceAPIVersionCache is a mapping of ResourceType to APIVersion
+	// so that we don't query supported API versions on each call to login for
+	// a given resource type
+	resourceAPIVersionCache map[string]string
 }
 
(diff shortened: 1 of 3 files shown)


backend.go
```diff
@@ -23,6 +23,11 @@ type azureAuthBackend struct {
 	l sync.RWMutex
 
 	provider provider
+
+	// resourceAPIVersionCache is a mapping of ResourceType to APIVersion
+	// so that we don't query supported API versions on each call to login for
+	// a given resource type
+	resourceAPIVersionCache map[string]string
```
fairclothjm, line 30: will have the following structure:
```
resourceAPIVersionCache: map[string]string [
    "Microsoft.Web/sites": "2022-03-01",
]
```

vinay-gopalan (approved):
LGTM!

PR #80, opened 2023-01-03
Replace usage of useragent.String with useragent.PluginString

austingebauer:
## Overview

This PR replaces usage of [`useragent.String`](https://github.com/hashicorp/vault/blob/main/sdk/helper/useragent/useragent.go#L43) with [`useragent.PluginString`](https://github.com/hashicorp/vault/blob/main/sdk/helper/useragent/useragent.go#L58). `useragent.String` has been deprecated. Using `useragent.PluginString` will allow plugins running external to Vault to use correct Vault version information in construction of user-agent request headers regardless of the compiled SDK version.

## Related Issues/Pull Requests
- https://github.com/hashicorp/vault/pull/14229

### DIFF
--- a/azure.go
+++ b/azure.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/coreos/go-oidc"
 	"github.com/hashicorp/go-cleanhttp"
+	"github.com/hashicorp/vault/sdk/helper/useragent"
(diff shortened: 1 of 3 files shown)


austingebauer:
Whoops, description is a bit off on this one 😄 This wasn't using `useragent.String`. I can close this if no one sees value in this change.

fairclothjm (approved):
Seems ok to merge since it will make it more similar to the other plugins using this method.

PR #81, opened 2023-01-31
Use lock for API version cache

fairclothjm:
Plugins should handle multiple concurrent requests from Vault. So we add a mutex to the API version cache to prevent data races.


austingebauer (approved):
👍

PR #88, opened 2023-02-01
Implement Rotate Root with `go-autorest`

vinay-gopalan:
# Overview
A high level description of the contribution, including:

This PR adds an implementation of the rotate root feature to the Azure Auth project. The code in this PR uses the current implementation of rotate root from Azure Secrets, and makes use of the `go-autorest` library to talk to the MS Graph API.

This change enables users to rotate the root credentials for their Azure application using Vault.

# Contributor Checklist
- [x] Relevant docs are being tracked in a separate PR upstream https://github.com/hashicorp/vault/pull/18780
- [x] Backwards compatible


path_rotate_root.go
```diff
@@ -0,0 +1,149 @@
+package azureauth
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/hashicorp/go-multierror"
+	//"github.com/hashicorp/go-uuid"
```
austingebauer, line 10: Any reason to not use `github.com/hashicorp/go-uuid`?

wal.go
```diff
@@ -0,0 +1,105 @@
...
+	walRotateRootCreds = "rotateRootCreds"
+)
+
+// Eventually expire the WAL if for some reason the rollback operation consistently fails
+var maxWALAge = 24 * time.Hour
+
+func (b *azureAuthBackend) walRollback(ctx context.Context, req *logical.Request, kind string, data interface{}) error {
+	switch kind {
+	case walAppKey:
+		return b.rollbackAppWAL(ctx, req, data)
```
austingebauer, line 23: We shouldn't be creating or rolling back applications in the auth method. Suggest to remove everything related to rolling back applications from this.

api/api.go
```diff
@@ -0,0 +1,45 @@
+package api
```
austingebauer, line 1: Hmm, it feels a little odd to have the MS graph client in an "api" package while the other clients are in the top-level "azureauth" package (azure.go file). I know the package structure is like this in Azure secrets, but I think we should do what's best to help with readability/comprehension in the context of this repo.

Open to ideas! One idea would be a "client" package to organize related client code. Another idea would be to have all client-related code in the azure.go file.

api/api.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/date"
+)
+
+// AzureProvider is an interface to access underlying Azure Client objects and supporting services.
+// Where practical the original function signature is preserved. Client provides higher
+// level operations atop AzureProvider.
+type AzureProvider interface {
+	MSGraphClient
+}
```
austingebauer, line 16: I wonder if we need the `AzureProvider` interface? I don't see the value it provides when it embeds `MSGraphClient` and only has a single implementation. I _think_ this would still compile if you removed these lines?

austingebauer (commented):
Thanks, @vinay-gopalan. Took a first pass on this. Will return to it again.

azure.go
```diff
@@ -288,6 +332,12 @@ func (b *azureAuthBackend) getAzureSettings(ctx context.Context, config *azureCo
 		return nil, errors.New("resource is required")
 	}
 
+	subscriptionID := os.Getenv("AZURE_SUBSCRIPTION_ID")
```
fairclothjm, line 335: What is subscriptionID doing for rotate root? If I don't set this via the env or config, or if I set it to a bogus value I can still call rotate root and get a success. Maybe I am doing something wrong?

vinay-gopalan: @fairclothjm my understanding was that the subscription ID needs to be valid in order to set up a new client that allows us to talk to the MS Graph API. This happens on [this line](https://github.com/hashicorp/vault-plugin-auth-azure/pull/88/files#diff-f59e54a2e0d262d172de754ffdd049bf159deec9c5e0896fce921c2a0603b2dcR56) of our MS Graph implementation

When my credentials were invalid, I was having trouble with setting up the client and with talking to the API... not sure how a bogus value is yielding a valid client 🤔 @austingebauer do you have any thoughts on this?

path_rotate_root.go
vinay-gopalan: Nope! Nice catch 😄 Updated in latest commit

wal.go
vinay-gopalan: Ah nice, that was a bit that was confusing me; thanks for clarifying! Removed all delete application logic from this flow (also simplified our interface 😅)

api/api.go
vinay-gopalan: I do like the idea of having all client implementations in a separate package for readability and maintenance. `azure.go` is already quite populated with the implementation and setup of the Azure Provider, so it would make it pretty big and make it harder to find things in that file if we move everything in there 😅 

I like the idea of a `client` package which separates all the clients from the Azure Provider interface and its setup, and I'm also open to thoughts from other folks on the team! :)

fairclothjm: I think having a client package and moving everything there makes the most sense. But maybe that should be scoped to another task? We could just rename `api` -> `client` for now.

austingebauer: +1 on @fairclothjm idea. We can rename the api package to client. Then we can create a task to move the rest of the client code into that package (after 1.13 when we have more time).

path_rotate_root_test.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if newConfig.ClientSecret != config.NewClientSecret {
+		t.Fatal(fmt.Errorf("old and new password aren't equal after periodic function, they should be"))
+	}
+}
+
+func TestRotateRootPeriodicFunctionBeforeMinute(t *testing.T) {
```
vinay-gopalan, line 96: Not sure of what the expected behavior of this test should be. The test passes and a new client credential is created for my azure application, however the periodic func doesn't actually remove the old application password. Would appreciate any feedback on what this test should actually do 🙏🏼

api/api.go
vinay-gopalan: Renamed the package to `client` and also renamed `api.go` -> `client.go`. Thanks for the feedback! 😄

backend.go
```diff
@@ -54,16 +59,100 @@ func backend() *azureAuthBackend {
...
+		PeriodicFunc: b.periodicFunc,
 	}
 
 	b.resourceAPIVersionCache = make(map[string]string)
 
-	return b
+	return &b
+}
+
+func (b *azureAuthBackend) periodicFunc(ctx context.Context, sys *logical.Request) error {
```
austingebauer, line 77: I think it's worth adding a comment here (maybe some other locations too?) to note that the periodic function is responsible for eventually swapping out the root credential for rotation operations.

path_rotate_root.go
```diff
@@ -0,0 +1,129 @@
+package azureauth
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-uuid"
+
```
austingebauer, line 10:
```suggestion
```

path_login_test.go
```diff
@@ -16,9 +16,12 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/msi/armmsi"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-01-01-preview/authorization"
 	"github.com/coreos/go-oidc"
 	"github.com/hashicorp/vault/sdk/helper/policyutil"
 	"github.com/hashicorp/vault/sdk/logical"
+
```
austingebauer, line 23:
```suggestion
```

azure.go
```diff
@@ -329,12 +392,11 @@ func (b *azureAuthBackend) getAzureSettings(ctx context.Context, config *azureCo
 
 func ConfigurationFromName(name string) (cloud.Configuration, error) {
 	configs := map[string]cloud.Configuration{
-		"AZURECHINACLOUD":        cloud.AzureChina,
-		"AZUREPUBLICCLOUD":       cloud.AzurePublic,
-		"AZUREUSGOVERNMENTCLOUD": cloud.AzureGovernment,
+		"AzureChinaCloud":        cloud.AzureChina,
+		"AzurePublicCloud":       cloud.AzurePublic,
+		"AzureUSGovernmentCloud": cloud.AzureGovernment,
```
austingebauer, line 397: Wondering if we changed normalizing the input here so that things match up with casing in `GetGraphURI()`? I sort of liked that users could supply it in any casing instead of requiring them to get the pascal case right. Any chance we could continue to allow users to supply it in any casing by modifying `GetGraphURI()` instead?

azure.go
```diff
@@ -317,6 +366,20 @@ func (b *azureAuthBackend) getAzureSettings(ctx context.Context, config *azureCo
...
+	}
+	if envName == "" {
+		settings.Environment = azure.PublicCloud
+	} else {
+		var err error
+		settings.Environment, err = azure.EnvironmentFromName(envName)
+		if err != nil {
+			return nil, err
+		}
+	}
```
austingebauer, line 381: I wonder if this can be combined with the `os.Getenv("AZURE_ENVIRONMENT")` above on L355? I understand that `settings.Environment` and `settings.CloudConfig` take different values, but it appears a little redundant to read the same env var twice.

path_rotate_root_test.go
fairclothjm: This test seems to be verifying that we don't do a rotation until `NewClientSecretCreated` is greater than 1 minute. See https://github.com/hashicorp/vault-plugin-auth-azure/pull/88/files#diff-f7febf2126e4c13f0730b2d06b8a6d3bdcb748f6379770c478888c2e4ea1949eR96

So I think it is ok, assuming that is the behavior that we want. Maybe we could add a comment above this test describing the purpose?

fairclothjm: You can check it by adding this log line [before that return nil](https://github.com/hashicorp/vault-plugin-auth-azure/pull/88/files#diff-f7febf2126e4c13f0730b2d06b8a6d3bdcb748f6379770c478888c2e4ea1949eR97)

```
b.Logger().Debug("periodic func", "rotate-root", "Password should be at least a minute old before we process it")
```
and then add another test case here
```
{
	Name:    "70 seconds test:",
	Created: time.Second * 70,
},

```

path_rotate_root.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		b.Logger().Error("rotate root", "delete wal", err)
+	}
+
+	return nil, err
+}
+
+type passwordRemover interface {
+	RemoveApplicationPassword(ctx context.Context, applicationObjectID string, keyID string) error
+}
```
austingebauer, line 116: I think this is another one of those interfaces we can remove. I don't foresee us having more than one implementation, so it doesn't provide much value.

azure.go
vinay-gopalan: Ah, good call. Simplified the logic a bit so we can combine the two operations

path_rotate_root.go
vinay-gopalan: Ah nice! Removed in latest commit

azure.go
vinay-gopalan: That makes sense; I updated `GetGraphURI()` to check on the uppercased string instead of the pascal case. This allows the user to supply the `environment` field whatever way they prefer

backend.go
vinay-gopalan: Added some documentation for the function and added comments in places that it is being used

path_config.go
```diff
@@ -178,9 +217,29 @@ func (b *azureAuthBackend) pathConfigDelete(ctx context.Context, req *logical.Re
...
+	// built using old versions of this data
+	b.reset()
+
+	return nil
+}
+
 const (
-	confHelpSyn  = `Configures the Azure authentication backend.`
-	confHelpDesc = `
+	defaultRootPasswordTTL = 4380 * time.Hour
```
austingebauer, line 239: I think it'd be good to preserve [this comment](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/path_config.go#L19-L21) from Azure secrets.

path_rotate_root_test.go
vinay-gopalan: Ah nice! Thanks for the clarification; I'll add a brief description to the test, because this confused me a bit 😅

vinay-gopalan: Added a docstring to the test to better describe its working. Thanks! 🙏🏼

path_config.go
vinay-gopalan: Added in latest commit

azure.go
```diff
@@ -304,17 +353,27 @@ func (b *azureAuthBackend) getAzureSettings(ctx context.Context, config *azureCo
...
 		settings.CloudConfig, err = ConfigurationFromName(configName)
 		if err != nil {
 			return nil, err
 		}
+
+		settings.Environment, err = azure.EnvironmentFromName(envName)
+		if err != nil {
+			return nil, err
+		}
 	}
```
austingebauer, line 377: `configName` and `envName` are the same values, so I think this can be simplified a bit:

```suggestion
	environment := os.Getenv("AZURE_ENVIRONMENT")
	if environment == "" {
		// set environment from config
		environment = config.Environment
	}
	if environment == "" {
		// use default values if no environment is provided
		settings.CloudConfig = cloud.AzurePublic
		settings.Environment = azure.PublicCloud
	} else {
		var err error
		settings.CloudConfig, err = ConfigurationFromName(environment)
		if err != nil {
			return nil, err
		}

		settings.Environment, err = azure.EnvironmentFromName(environment)
		if err != nil {
			return nil, err
		}
	}
```

wal.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+const (
+	walRotateRootCreds = "rotateRootCreds"
+)
+
+// Eventually expire the WAL if for some reason the rollback operation consistently fails
+var maxWALAge = 24 * time.Hour
```
austingebauer, line 16: Unused?

backend.go
```diff
@@ -54,16 +59,105 @@ func backend() *azureAuthBackend {
...
-				pathLogin(b),
-				pathConfig(b),
+				pathLogin(&b),
+				pathConfig(&b),
+				pathRotateRoot(&b),
 			},
-			pathsRole(b),
+			pathsRole(&b),
 		),
+		WALRollback: b.walRollback,
```
austingebauer, line 68: Azure secrets sets the [WALRollbackMinAge](https://github.com/hashicorp/vault-plugin-secrets-azure/blob/main/backend.go#L70-L72) to 10 minutes. I don't think we need it to be 10 minutes here, but I do think we should set this value. Otherwise, the rollback function could try to rollback a root credential rotation that's currently in flight.

I'm thinking 3 minutes is sufficient. That gives the periodic func at least 1 attempt to complete the rotation.

azure.go
vinay-gopalan: Updated, thanks!

wal.go
vinay-gopalan: Yup 😅 removed in latest commit

backend.go
vinay-gopalan: Added a min age of 3 minutes to the wal rollback 🙏🏼

path_rotate_root_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+// This test performs a rotate root operation and
+// swaps the root credentials for the application
+func TestRotateRootSuccess(t *testing.T) {
+	b, s := getTestBackend(t)
+
+	subscriptionID, tenantID, clientID, clientSecret := getAzureEnvironmentSettings()
+	if subscriptionID == "" || tenantID == "" ||
+		clientID == "" || clientSecret == "" {
+		t.Skip("environment variables not set, skipping test in CI")
```
austingebauer, line 21: Nit: Was a little confused by "in CI" here when I ran this locally. Suggest to remove this mention since it can be ran outside of CI context as well.

azure.go
vinay-gopalan: @fairclothjm, @austingebauer and I both did some digging into the implementation of the client in the `authorization` package, and it looks like the subscription ID is in fact not necessary to perform the rotate root operation, nice catch! Removed the extra bits of code that add the subscription ID. Since the sub ID is never being used, we are now simply passing an empty string into the `NewWithBaseURI` function [here](https://github.com/hashicorp/vault-plugin-auth-azure/pull/88/files#diff-6bf5bfdeab8247e592e67fdaf0494d289ff541b16cc55fad81542cbfdef15f1eR47)

path_rotate_root_test.go
vinay-gopalan: Good point, fixed error messaging in latest commit

client/application_msgraph.go
```diff
@@ -0,0 +1,242 @@
...
+		return "https://microsoftgraph.chinacloudapi.cn", nil
+	default:
+		return "", fmt.Errorf("environment '%s' unknown", env)
+	}
+}
+
+// NewMSGraphApplicationClient creates a new MS Graph Client
+// to allow interaction with the MS Graph v1.0 API.
+func NewMSGraphApplicationClient(userAgentExtension string, graphURI string, auth autorest.Authorizer) (*AppClient, error) {
+	client := authorization.NewWithBaseURI(graphURI, "")
```
austingebauer, line 47: I think it'd be useful to add a comment here explaining that we're intentionally not providing the subscription ID since it's not needed for our usage of the client.

client/application_msgraph.go
```diff
@@ -0,0 +1,242 @@
...
+	Value []ApplicationResult `json:"value"`
+}
+
+func GetGraphURI(env string) (string, error) {
+	switch strings.ToUpper(env) {
+	case "AZUREPUBLICCLOUD", "":
+		return "https://graph.microsoft.com", nil
+	case "AZUREUSGOVERNMENTCLOUD":
+		return "https://graph.microsoft.us", nil
+	case "AZUREGERMANCLOUD":
```
austingebauer, line 35: I'm thinking we should keep the cloud envs consistent between `GetGraphURI` and `ConfigurationFromName` wrt AzureGermanCloud. I remember it being shut down from https://github.com/hashicorp/vault-plugin-auth-azure/pull/72#discussion_r1043869392, so we should remove it here.

vinay-gopalan: Updated in latest commits

client/application_msgraph.go
vinay-gopalan: Added a comment to document the lack of subscription ID

go.mod
```diff
@@ -3,62 +3,78 @@ module github.com/hashicorp/vault-plugin-auth-azure
...
 	github.com/hashicorp/vault/api v1.8.3
 	github.com/hashicorp/vault/sdk v0.7.0
 	golang.org/x/oauth2 v0.4.0
 )
 
-require github.com/hashicorp/errwrap v1.1.0 // indirect
+require (
+	github.com/hashicorp/errwrap v1.1.0 // indirect
+	github.com/mitchellh/mapstructure v1.5.0 // indirect
+)
```
austingebauer, line 26: Hmm, I think you can manually move these two into the require block below and then run `go mod tidy`. Not sure why they're split out 🤔

austingebauer (approved):
LGTM! Thanks for all the work on this!

go.mod
vinay-gopalan: Fixed! 🙏🏼

fairclothjm (approved):
LGTM!

PR #96, opened 2023-02-22
enable multiplexing

raymonstah:
# Overview

* the plugin will be multiplexed when run as an external plugin against vault versions that support plugin multiplexing
* we continue to set the TLSProviderFunc to maintain backwards compatibility with vault versions that don't support AutoMTLS (< 1.12)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 CHANGES:
 
-ADD YOUR CHANGE HERE [GH-XXX]
+* ADD YOUR CHANGE HERE [GH-XXX]
 
 IMPROVEMENTS:
 
-Enable multiplexing [GH-172]
+* Enable multiplexing [[GH-96](https://github.com/hashicorp/vault-plugin-auth-azure/pull/96)]
+
 ## v0.13.0
 
 FEATURES:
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -15,11 +15,13 @@ require (
...
 	github.com/hashicorp/go-hclog v1.4.0
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/hashicorp/vault/api v1.8.3
-	github.com/hashicorp/vault/sdk v0.7.0
+	github.com/hashicorp/vault/api v1.9.0
+	github.com/hashicorp/vault/sdk v0.8.1
 	golang.org/x/oauth2 v0.4.0
 )
 
+require github.com/hashicorp/go-kms-wrapping/v2 v2.0.7 // indirect
```
vinay-gopalan, line 23: We could probably add this to the `require` block below and run `go mod tidy`. Not sure why Go added this separately 🤔

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Unreleased
 
+CHANGES:
+
+* ADD YOUR CHANGE HERE [GH-XXX]
```
vinay-gopalan, line 5: I think we can skip the `CHANGES` section since we only need the Improvements

vinay-gopalan (approved):
Looks good! Had a couple of nits/small comments but other than that everything looks great 👍🏼

fairclothjm (approved):

PR #100, opened 2023-03-02
Use application password instead of service principal password while bootstrapping environment

vinay-gopalan:
This PR updates the bootstrap TF file and sets up an Azure Application Password instead of a Service Principal Password. While the SP Password is virtually the same as the App Password, it does not show up in the Certificates and Secrets section in the Azure Portal.


austingebauer (approved):
Thanks, @vinay-gopalan!

PR #107, opened 2023-04-28
Use MSGraph SDK and deprecate Autorest for rotate root operations

vinay-gopalan:
This PR deprecates the final bits of `go-autorest` from the codebase as that package has stopped receiving updates, and ports the MS Graph Application operations over to the new MSGraph SDK for Go.


austingebauer:
Closing in favor of https://github.com/hashicorp/vault-plugin-auth-azure/pull/121.

PR #121, opened 2023-07-19
Replaces go-autorest MS graph client with msgraph-sdk-go

austingebauer:
## Overview

This PR replaces the deprecated [go-autorest](https://github.com/Azure/go-autorest) module with [msgraph-sdk-go](https://github.com/microsoftgraph/msgraph-sdk-go). Additionally, it migrates the [azure-sdk-for-go](https://github.com/Azure/azure-sdk-for-go/blob/main/documentation/MIGRATION_GUIDE.md) to the new package structure (see related [issue](https://github.com/hashicorp/vault-plugin-secrets-azure/issues/143)).

There are no known API changes or compatibility breaks in this PR.

## Testing

All tests in the repository are passing. Additionally, I manually tested the secrets engine functionality using our Terraform [bootstrap](https://github.com/hashicorp/vault-plugin-auth-azure/tree/main/bootstrap/terraform) which pulls an [IMDS token](https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token) off of a VM.

I've additionally added an acceptance test that actually uses Azure resources to login in.

## Related Issues/Pull Requests
- https://github.com/hashicorp/vault-plugin-secrets-azure/issues/143

## Contributor Checklist

- [x] Backwards compatible


azure.go
```diff
@@ -32,38 +29,21 @@ import (
...
-}
-
-type providersClient interface {
-	Get(ctx context.Context, resourceProviderNamespace string, options *armresources.ProvidersClientGetOptions) (armresources.ProvidersClientGetResponse, error)
-}
-
-type tokenVerifier interface {
-	Verify(ctx context.Context, token string) (*oidc.IDToken, error)
-}
+// https://learn.microsoft.com/en-us/graph/sdks/national-clouds
```
austingebauer, line 32: Couldn't find constants for these in the library code.

azure.go
```diff
@@ -391,3 +361,19 @@ func ConfigurationFromName(name string) (cloud.Configuration, error) {
 
 	return c, nil
 }
+
+func graphURIFromName(name string) (string, error) {
+	configs := map[string]string{
+		"AZURECHINACLOUD":        azureChinaCloudBaseURI,
```
raymonstah, line 367: it might be worthwhile to make these constants as well

raymonstah (approved):
Looks good! question about the error handling: I noticed that a lot of the `if err` checks return the error directly. Do we feel that the error returned provides enough context? If not, we can consider wrapping them.

azure.go
austingebauer: Good idea. Done in https://github.com/hashicorp/vault-plugin-auth-azure/pull/121/commits/2b2bf18974cc94962f388ff8a574c9579925dd90.

austingebauer:
Good point @raymonstah. Checking some locations to see if we can provide more error context.

vinay-gopalan (approved):
Looks great, thanks for working through this!

PR #122, opened 2023-07-20
Add WIF support

kpcraig:
# Overview
This PR adds a check against user-assigned identity ClientIDs iff the PrincipalID fails to validate. This should allow WIF-based AKS agents to authenticate via vault.

If you look at the code, you might notice we grab an MSIClient for a second possible time to search through the list of managed identities in our resource group. It sure feels like there should be a way to do what I want to do during the steps where we enumerate the possible PrincipalIDs, but I couldn't figure out the right incantation. It might also cause extra work, since we really only want to do the AppID check if the ObjectID check fails.

In order to eliminate a janky type assertion that existed in an earlier version of this code, I added a the MSIClient function I wanted to use to the MSI Client interface defined in `azure.go`. This also means I can override the function in the mock, although since the function returns a `runtime.Pager`, it's a bit more complicated than the override of the `Get` function that already exists.


path_login.go
```diff
@@ -176,6 +180,7 @@ func (b *azureAuthBackend) pathLogin(ctx context.Context, req *logical.Request,
 			Metadata: map[string]string{
 				"resource_group_name": resourceGroupName,
 				"subscription_id":     subscriptionID,
+				"app_id":              appID,
```
austingebauer, line 183: Curious if there is any particular reason to put the app ID into the alias and token metadata?

azure.go
```diff
@@ -42,6 +43,7 @@ type vmssClient interface {
 
 type msiClient interface {
 	Get(ctx context.Context, resourceGroupName string, resourceName string, options *armmsi.UserAssignedIdentitiesClientGetOptions) (armmsi.UserAssignedIdentitiesClientGetResponse, error)
+	NewListByResourceGroupPager(resourceGroupName string, options *armmsi.UserAssignedIdentitiesClientListByResourceGroupOptions) *runtime.Pager[armmsi.UserAssignedIdentitiesClientListByResourceGroupResponse]
```
austingebauer, line 46: 👍

path_login.go
```diff
@@ -258,7 +264,7 @@ func (b *azureAuthBackend) verifyClaims(claims *additionalClaims, role *azureRol
 	return nil
 }
 
-func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, resourceGroupName, vmName, vmssName, resourceID string, claims *additionalClaims, role *azureRole) error {
+func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, resourceGroupName, vmName, vmssName, resourceID, appID string, claims *additionalClaims, role *azureRole) error {
```
austingebauer, line 267: `appID` unused?

path_login.go
```diff
@@ -401,7 +407,33 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 	// Ensure the token OID is the principal id of the system-assigned identity
 	// or one of the user-assigned identities
 	if _, ok := principalIDs[claims.ObjectID]; !ok {
```
austingebauer, line 409: Curious if we still require one of `vm_name`, `vmss_name`, or `resource_id` to be provided? Do you know if it still makes sense for that requirement to be true in the WIF environment?

path_login.go
```diff
@@ -401,7 +407,33 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
-		return errors.New("token object id does not match expected identities")
+		// if it isn't, check the appID and see if _that_ exists.
```
austingebauer, line 410: Suggest to expand on this comment to explain any non-obvious details that might be easy to forget in the future. For example, it might be worth mentioning that this fallback validation is in place to support WIF and touch on what's being validated.

path_login_test.go
```diff
@@ -643,6 +643,59 @@ func TestLogin_BoundScaleSet(t *testing.T) {
 	testLoginFailure(t, b, s, loginData, claims, roleData)
 }
 
+func TestLogin_AppID(t *testing.T) {
```
austingebauer, line 646: Thanks for adding this!

austingebauer (commented):
Looking good! Left a few small Qs and suggestions.

path_login.go
kpcraig: I believe you're right, we should check app_id if authentication vial vm/vmss/resource_id fails *or* if no such identifier was provided.

path_login.go
kpcraig: added some words!

path_login.go
kpcraig: I believe the idea was that it could be part of the authorization decision now, so put it in there. I don't think we _have_ to have it?

I did refactor it a bit so that it should only set if present in the claim.

path_login.go
kpcraig: i tweaked the switch branches a bit for this

azure.go
kpcraig: I was torn, but committing to this did make writing the mocked-out test case much easier.

path_login.go
austingebauer: We don't have to include it, but I like the idea of including it because it's a part of the authorization decision. I'd say let's leave it. Good idea to conditionally add it.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Unreleased
 
+FEATURES:
+* Allow WIF based Azure resources to authenticate with Vault
```
austingebauer, line 4: Nit but might spell out the acronym here?

path_login.go
```diff
@@ -396,12 +398,48 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
 	}
 
 	// Ensure the token OID is the principal id of the system-assigned identity
 	// or one of the user-assigned identities
 	if _, ok := principalIDs[claims.ObjectID]; !ok {
-		return errors.New("token object id does not match expected identities")
+		// if it isn't, check the appID and see if _that_ exists. In some cases, particularly WIF (workload identity
+		// federation), there is no principal that matches the incoming ObjectID. In this case, we can still validate
+		// by checking the appID against the list of managed identities. (The appID is valid for use with authorizing
+		// claims, per https://learn.microsoft.com/en-us/azure/active-directory/develop/access-tokens#payload-claims)
```
austingebauer, line 416: 👍

path_login.go
```diff
@@ -396,12 +398,48 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 		for _, userIdentity := range resp.Identity.UserAssignedIdentities {
 			principalIDs[convertPtrToString(userIdentity.PrincipalID)] = struct{}{}
 		}
+	default:
+		// in some cases (particularly WIF), a vm/vmss/resource_id might not be provided, in that case
+		// we'll try to authenticate by matching the claim's app_id to the list of managed identities
+		// (see the comment below on that)
+		if claims.AppID == "" {
+			return errors.New("one of vm_name, vmss_name, resource_id, or an app_id JWT claim must be provided")
```
austingebauer, line 406: I think the app ID claim key is `appid` instead of `app_id`?

austingebauer (approved):
Nice job! LGTM.

path_login.go
kpcraig: it is, ty!

path_login.go
```diff
@@ -396,12 +398,48 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
 	// or one of the user-assigned identities
 	if _, ok := principalIDs[claims.ObjectID]; !ok {
-		return errors.New("token object id does not match expected identities")
+		// if it isn't, check the appID and see if _that_ exists. In some cases, particularly WIF (workload identity
+		// federation), there is no principal that matches the incoming ObjectID. In this case, we can still validate
+		// by checking the appID against the list of managed identities. (The appID is valid for use with authorizing
+		// claims, per https://learn.microsoft.com/en-us/azure/active-directory/develop/access-tokens#payload-claims)
+		if claims.AppID != "" {
+			clientIDs := map[string]struct{}{}
+			c, err := b.provider.MSIClient(subscriptionID) // this is the second time we're calling this, is there a way to reuse that?
```
calvn, line 419: IIUC the other place within this method where this gets called is if vmss is provided, but that is not guaranteed. So the call here might be necessary.

path_login.go
```diff
@@ -396,12 +398,48 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
+					if identity.Properties != nil && identity.Properties.ClientID != nil {
+						clientIDs[*identity.Properties.ClientID] = struct{}{}
+					}
+				}
+			}
+
+			if _, ok := clientIDs[claims.AppID]; !ok {
+				return errors.New("neither token object id nor token app id match expected identities")
+			}
+		} else {
```
calvn, line 439: Nit: You could probably switch the statement so we we check for `if claims.AppID = "" { return error }` first so the happy path doesn't get this extra nesting.

calvn (approved):
Left a small nit + comment cleanup

path_login_test.go
```diff
@@ -643,6 +643,59 @@ func TestLogin_BoundScaleSet(t *testing.T) {
...
+	}
+
+	loginData := map[string]interface{}{
+		"role": roleName,
+	}
+	testLoginFailure(t, b, s, loginData, claims, roleData)
+
+	loginData["resource_group_name"] = resourceGroup
+	loginData["subscription_id"] = "sub"
+	//loginData["resource_id"] = "/subscriptions/eb936495-7356-4a35-af3e-ea68af201f0c/resourceGroups/azure-func-rg/providers/Microsoft.Web/sites/my-azure-func"
```
calvn, line 691: Extraneous comment here?

path_login.go
kpcraig: yeah I was thinking i could make the MSIClient run universal, but it might require protection via if statement, which doesn't actually make the code clearer, so it might be for the best.

path_login.go
kpcraig: good call!

PR #127, opened 2023-08-04
VAULT-16420: Fix for ClientID in different resource group

kpcraig:
This PR adds code that will scan through a role's bound resource groups as well as the resource group in the request to see if an AppID matches.

This should resolve an issue where a service inside an AKS gets the 'AKS resource group' while the Application identity is in a different (e.g., the parent) resource group.

When a match of this type occurs, we also skip the bound resource group check, since a bound resource group match is implied by the IDs existence.

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -407,6 +407,7 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 		}
 	}
 
+	var wifMatch bool
 	// Ensure the token OID is the principal id of the system-assigned identity
 	// or one of the user-assigned identities
 	if _, ok := principalIDs[claims.ObjectID]; !ok {
@@ -423,6 +424,8 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 		if err != nil {
 			return fmt.Errorf("failed to create client to retrieve app ids: %w", err)
 		}
+
+		// add the identities associated with this resource group
 		pager := c.NewListByResourceGroupPager(resourceGroupName, &armmsi.UserAssignedIdentitiesClientListByResourceGroupOptions{})
 		for pager.More() {
 			page, err := pager.NextPage(ctx)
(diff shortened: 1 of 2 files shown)


VioletHynes (approved):
Looks good! I think this should do the trick.

path_login.go
```diff
@@ -436,18 +439,35 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
 			}
 		}
 
+		// add the identities associated with the bound resource groups, if any
+		for _, rg := range role.BoundResourceGroups {
+			pager := c.NewListByResourceGroupPager(rg, &armmsi.UserAssignedIdentitiesClientListByResourceGroupOptions{})
```
fairclothjm, line 444: Nit: could we either refactor this common code into a helper function or append `resourceGroupName` to this array?

path_login.go
```diff
@@ -436,18 +439,35 @@ func (b *azureAuthBackend) verifyResource(ctx context.Context, subscriptionID, r
...
 	}
 
 	// Check bound subscriptions
 	if len(role.BoundSubscriptionsIDs) > 0 && !strListContains(role.BoundSubscriptionsIDs, subscriptionID) {
 		return errors.New("subscription not authorized")
 	}
 
-	// Check bound resource groups
-	if len(role.BoundResourceGroups) > 0 && !strListContains(role.BoundResourceGroups, resourceGroupName) {
+	// Check bound resource groups unless we matched due to WIF (TODO: explanation for why we do this)
```
fairclothjm, line 469: I like the idea of adding more details on why we short circuit for WIF.

fairclothjm (approved):
LGTM! Just some suggestions but nothing blocking

path_login.go
kpcraig: yeah i think appending it would work

fairclothjm (approved):

PR #133, opened 2023-09-12
Add Azure API retry configurables 

### LINKED ISSUE #123: Expose client retry options as a config knob
The current [azure client initialization](https://github.com/hashicorp/vault-plugin-auth-azure/blob/main/azure.go#L236) does not allow for customizations of the retry behavior that is available in the[ underlying sdk](https://github.com/Azure/azure-sdk-for-go/blob/c83d84e92caa9c68d6ad90d5158b395d97b3532d/sdk/azcore/runtime/policy_retry.go#L29-L34).

We're currently experiencing extreme latencies from the ARM API and not being able to fine tune settings like this exacerbates the problem. Please consider exposing the retry behavior similar to the [aws auth config knob](https://github.com/hashicorp/vault/blob/main/builtin/logical/aws/path_config_root.go#L48-L52).

jasonodonnell:
This allows operators to configure the retry behavior when API failures occur. The defaults hardcoded here reflect the defaults in the Azure SDK. 

Fixes #123.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+IMPROVEMENTS:
+    * Added Azure API configurable retry options [GH-133](https://github.com/hashicorp/vault-plugin-auth-azure/pull/133)
+
 ## v0.16.1
 IMPROVEMENTS:
 * Updated dependencies:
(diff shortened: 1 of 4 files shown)


maxcoulombe (approved):
LGTM

austingebauer (approved):
Thanks again for picking this up, @jasonodonnell!

PR #139, opened 2023-11-02
Remove dependency on github.com/microsoftgraph/msgraph-sdk-go

swenson:
We only use msgraph-sdk-go for a handful of calls, and they are relatively straightforward HTTP calls.

This code still uses the Microsoft packages for authentication, which is the trickier part.

I'm having trouble getting the tests to pass consistently, but I always have trouble with these tests.

I compiled the Vault binary with this change only, and it reduces the size of the Vault binary from 360 MB to 264 MB.

### DIFF
--- a/client/msgraph.go
+++ b/client/msgraph.go
@@ -4,28 +4,83 @@
 package client
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/google/uuid"
-	msgraphsdkgo "github.com/microsoftgraph/msgraph-sdk-go"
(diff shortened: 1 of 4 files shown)


client/msgraph.go
```diff
@@ -41,63 +96,138 @@ func NewMSGraphApplicationClient(graphURI string, creds azcore.TokenCredential)
...
+	}
+	request := &http.Request{
+		Method: "GET",
+		URL:    u,
+		Header: headers,
+	}
 
-	apps := resp.GetValue()
-	if len(apps) == 0 {
+	resp, err := http.DefaultClient.Do(request)
```
fairclothjm, line 130: Should we use https://github.com/hashicorp/go-cleanhttp?

swenson: Yep! I rushed this out, but if we intended to go this route, I'd refactor this to use a common http client and DRY a few things.

But, it sounds like we're looking to use the msgraph SDK in more places, so merging this wouldn't really change anything. I'll close for now but keep the branch.

PR #151, opened 2024-04-26
support workload identity federation

fairclothjm:
This PR adds plugin WIF support to the Azure auth method. The following new fields are added to the config endpoint to enable configuring Workload Identity Federation:

- identity_token_audience
- identity_token_ttl


bootstrap/terraform/vm.tf
```diff
@@ -111,7 +111,7 @@ resource "azurerm_public_ip" "vault_azure_pub_ip" {
 
 # Restrict SSH access to the local public IP
 data "http" "my_ip" {
-  url = "https://ifconfig.me/ip"
+  url = "https://ipinfo.io/ip"
```
thyton, line 114: For my learning purpose, could you share we made this change?

fairclothjm: I was unable to ssh into the Azure VM. "https://ifconfig.me/ip" returns an IPv6 and switching to "https://ipinfo.io/ip" resolves the issue. It returns IPv4. Maybe the original "https://ifconfig.me/ip" endpoint changed its behavior at some point?

thyton (approved):
LGTM. Thank you for reorganizing and refactoring the tests!

azure.go
```diff
@@ -330,6 +380,8 @@ func (b *azureAuthBackend) getAzureSettings(ctx context.Context, config *azureCo
 	}
 	settings.ClientSecret = clientSecret
 
+	settings.IdentityTokenAudience = config.IdentityTokenAudience
+
```
Zlaticanin, line 384: One small nit, we probably should set the token ttl here as well
```suggestion
settings.IdentityTokenTTL = config.IdentityTokenTTL
```

Zlaticanin (approved):
I went through the setup and was able to successfully use the WIF flow. Great job! 👏 🎉

azure.go
vinay-gopalan: Resolving since this was added in as part of the context removal PR!

vinay-gopalan (approved):
LGTM!

PR #153, opened 2024-05-13
add basic field validation on login path

kpcraig:
This PR adds some character validation to the login route, to ensure various values match azure's naming specifications.

In the future it might be worth discussing another "Validation" field to the framework package to allow us to specify validation characteristics directly, although this would be an SDK change.


thyton (approved):

azure.go
```diff
@@ -447,3 +448,10 @@ func graphURIFromName(name string) (string, error) {
 
 	return c, nil
 }
+
+var rx = regexp.MustCompile(`[a-zA-Z][a-zA-Z0-9\-]+`)
+
+// verify the field provided matches Azure's requirements.
```
fairclothjm, line 454: Should we provide a link to Azure's documented rules?

path_login.go
```diff
@@ -136,6 +136,19 @@ func (b *azureAuthBackend) pathLogin(ctx context.Context, req *logical.Request,
 	vmName := data.Get("vm_name").(string)
 	resourceID := data.Get("resource_id").(string)
 
+	if !validateAzureField(subscriptionID) {
+		return logical.ErrorResponse(fmt.Sprintf("invalid subscription id %q", subscriptionID)), nil
+	}
+	if !validateAzureField(resourceGroupName) {
```
fairclothjm, line 142: It looks like the regex is too restrictive based on the allowed characters listed here for resourceGroups: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftresources

path_login.go
```diff
@@ -136,6 +136,19 @@ func (b *azureAuthBackend) pathLogin(ctx context.Context, req *logical.Request,
 	vmName := data.Get("vm_name").(string)
 	resourceID := data.Get("resource_id").(string)
 
+	if !validateAzureField(nameRx, subscriptionID) {
```
fairclothjm, line 139: Where did you find the rules for subscription ID?

azure.go
```diff
@@ -447,3 +448,12 @@ func graphURIFromName(name string) (string, error) {
 
 	return c, nil
 }
+
+var nameRx = regexp.MustCompile(`^([a-zA-Z][a-zA-Z0-9\-]*)$|^$`)
```
fairclothjm, line 452: From what I can tell, a typical subscription ID looks like `1234abcd-1234-1234-abcd-12345abcdefg` which this regex will fail to match https://go.dev/play/p/lVWlfZNc42w

kpcraig: You're right, i was just being sloppy, it's "pretty obviously" a guid, although documentation is a bit thin on the ground stating it directly.

I was able to find this: https://learn.microsoft.com/en-us/rest/api/defenderforcloud/tasks/get-subscription-level-task

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 IMPROVEMENTS:
+* Add login field validation for subscription id, resource group name, vmss name, and vm name
```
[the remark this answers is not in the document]
kpcraig: I think it should be fixed now!

path_login_test.go
```diff
@@ -753,14 +753,62 @@ func TestLogin_AppID(t *testing.T) {
...
+	loginData["vm_name"] = "vm"
+	loginData["resource_group_name"] = "a/../b"
+	testLoginFailure(t, b, s, loginData, claims, roleData)
+
+	//claims["nbf"] = time.Now().Add(60 * time.Second).Unix()
+	//testLoginFailure(t, b, s, loginData, claims, roleData)
+	//
+	//claims["nbf"] = time.Now().Add(-60 * time.Second).Unix()
+	//claims["exp"] = time.Now().Add(-60 * time.Second).Unix()
+	//testLoginFailure(t, b, s, loginData, claims, roleData)
```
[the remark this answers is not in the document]
kpcraig: removed - believe this was extra from when i copy pasted code from another test to match the structure

fairclothjm (approved):

PR #168, opened 2024-10-11
fix: errcheck for parsing CLI flags

thevilledev:
# Overview

Add missing error check to `main()` for parsing CLI flags.

# Design of Change

Error check was missing.

None

# Contributor Checklist

Fixed in various Vault plugin repositories before such as https://github.com/hashicorp/vault-plugin-auth-kubernetes/pull/202 and https://github.com/hashicorp/vault/pull/28692


kpcraig (approved):
Thanks for the tidy!

PR #172, opened 2024-11-05
expand regex to allow underscores (and periods)

kpcraig:
This expands the "nameRx" - used for vm names and vmss names to allow underscores and periods. Disallows periods and hyphens from being the last character, as described by https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.VM.Name/#description

I also added parens to resource group names, since they seem to be allowed both in https://azure.github.io/PSRule.Rules.Azure/en/rules/Azure.ResourceGroup.Name/ and https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules (grep for 'resourcegroup')

### DIFF
--- a/azure.go
+++ b/azure.go
@@ -450,8 +450,9 @@ func graphURIFromName(name string) (string, error) {
 }
 
 // guidRx from https://learn.microsoft.com/en-us/rest/api/defenderforcloud/tasks/get-subscription-level-task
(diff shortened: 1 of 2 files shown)


tomcf-hcp:
I think the "allow" regex is easier to match the documentation, and less likely to cause an issue. Thanks for adding comments and references to the regex to indicate their intent and source, makes it easier to validate.

tomcf-hcp (approved):
Looks good to me, the regex's seem to match the rules.
Would like to know if JM has any other comments.

fairclothjm (approved):

PR #176, opened 2024-11-19
backport/1.16.x: bump deps to avoid old go-jose

kpcraig:
Bump oidc to v3 and go-jose to v.3.0.3 to remove dependency on go-jose 2.6.0 and go-jose 3.0.0

If the major version bump in oidc sketches you out, we also made the change in a "chore" update into 0.18.0, and there's no go-oidc v2 without an old go-jose (i.e., go-oidc v2.2.1 is the latest v2)

If the `.go-version` bump sketches you out, this both fixes an issue with the test checks, and matches the version used in the release/1.16.x branch of vault.


go.mod
```diff
@@ -1,19 +1,21 @@
 module github.com/hashicorp/vault-plugin-auth-azure
 
-go 1.20
+go 1.21
+
+toolchain go1.22.2
```
fairclothjm, line 5:
```suggestion
```

fairclothjm:
I altered the PR title slightly for posterity

PR #182, opened 2025-02-10
dependency updates

kpcraig:
Update dependencies before release


tvoran:
We'll probably want to update go in `.go-version` too, since that's the version used to build the binary release.

.go-version
```diff
@@ -1 +1 @@
-1.22.8
+1.23.3
```
tvoran, line 1: I think 1.23.3 has CVEs associated with it; I'd suggest going with the latest 1.23.

tvoran, line 1:
```suggestion
1.23.6
```

tvoran (dismissed):

hashiblaum (approved):

fairclothjm (approved):

PR #186, opened 2025-02-26
require resource_group_name , vm_name, and vmss_name to match token claims on login

thyton:
# Overview
Our API docs on the login endpoint says 👇 

> [resource_group_name](https://developer.hashicorp.com/vault/api-docs/auth/azure#resource_group_name) (string: <required>) - The resource group for the machine that generated the MSI token. This information can be obtained through instance metadata.
> [vm_name](https://developer.hashicorp.com/vault/api-docs/auth/azure#vm_name) (string: "") - The virtual machine name for the machine that generated the MSI token. This information can be obtained through instance metadata. If vmss_name is provided, this value is ignored.
> [vmss_name](https://developer.hashicorp.com/vault/api-docs/auth/azure#vmss_name) (string: "") - The virtual machine scale set name for the machine that generated the MSI token. This information can be obtained through instance metadata.

However, there is no binding between the JWT and vm_name / vmss_name / resource_group_name in the login handler, which can lead to multiple exploit scenarios. Both 1 and 2 are the scenarios that allow to bypass `bound_locations`.
1. Token of a VM named "test-vm-bad" is used to login with `vm_name`="test-vm-good"
2. Token of a VMSS named "test-vmss-bad" is used to login with `vmss_name`="test-vmss-good"
3. Token of a resource in a resource group named "test-rg-bad" is used to login with `resource_group_name`="test-rg-good"

The PR adds validations on the JWT claims against the provided vm_name / vmss_name / resource_group_name. 

## Integration test outputs
https://docs.google.com/document/d/1OdUjG9MBezNRoEom5uY_KQyVDicKi9Q2aWHaXCi1Gu8/

- [ ] [Issue #1234](https://github.com/hashicorp/vault/issues/1234)
- [ ] [PR #1234](https://github.com/hashicorp/vault/pr/1234)

# Contributor Checklist
- [ ] Backwards compatible

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -27,9 +27,21 @@ var defaultResourceClientAPIVersion = "2022-03-01"
 
 const (
 	// Depending on the identities are attached to a standalone or VMSS-stemmed virtual machine,
-	// we might get different claims in xms_mirid. If there is a user-assigned managed identity,
-	// the xms_mirid will be in the format of fmtRID. Otherwise, it will be in the format of
-	// fmtRIDWithUserAssignedIdentities, xms_az_rid will be in the format of fmtRID.
+	// we might get different claim patterns in xms_mirid.
+	// If system assigned managed identity is enabled and no identity is specified in the Instance
+	// Metadata Service request, xms_mirid will be in the format of fmtRID, and xms_az_rid doesn't exist.
+	// If system assigned managed identity is not enabled, and only one user-assigned identity exists,
+	// xms_mirid will be in the format of fmtRIDWithUserAssignedIdentities, xms_az_rid will be
+	// in the format of fmtRID.
+	// If system assigned managed identity is not enabled, and multiple user-assigned identities exist,
+	// users are required to specify a managed identity in the Instance Metadata Service request.
+	// xms_mirid will be in the format of fmtRIDWithUserAssignedIdentities, xms_az_rid will be
+	// in the format of fmtRID.
+	//
+	// See the following for more details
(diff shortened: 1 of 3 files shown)


path_login.go
```diff
@@ -25,6 +25,37 @@ import (
 // https://learn.microsoft.com/en-us/rest/api/resources/providers/get?tabs=HTTP
 var defaultResourceClientAPIVersion = "2022-03-01"
 
+const (
+	// Depending on the identities are attached to a standalone or VMSS-stemmed virtual machine,
+	// we might get different claims in xms_mirid. If there is a user-assigned managed identity,
```
fairclothjm, line 30: Should this say "system-assigned"?

fairclothjm (dismissed):
Great work @thyton!

path_login.go
thyton: Thanks JM! I updated the comment section with more details. Let me know if you have any suggestion!

fairclothjm (approved):

tonicmuroq:
I got metadata from a vmss within an AKS cluster, replaced sensitive data with `{{}}`:

```
"xms_az_rid": "/subscriptions/{{uuid}}/resourcegroups/{{rg}}/providers/Microsoft.Compute/virtualMachineScaleSets/{{vmss_name}}",
```

it's not just like `virtualMachines/%s`, but `virtualMachineScaleSets/%s`, i think this validation is not working on this case, or maybe i'm using this wrong? I'm actually running this application in AKS cluster, it's somehow different from running in azure virtual machines, how should I write my code to make this right in AKS pod?

PR #202, opened 2025-04-24
First pass at tests with implementation

wilsonge:
# Overview
As part of the hardening measures in PR 186 it didn't take into account azure virtual machine scalesets run in uniform mode - and only correctly validates scalesets deployed into flexible mode. This PR adds the code to cover both cases with appropriate tests.

- [x] [PR #186](https://github.com/hashicorp/vault-plugin-auth-azure/pull/186)

# Contributor Checklist
- [x] Backwards compatible

### DIFF
--- a/path_login.go
+++ b/path_login.go
@@ -45,6 +45,9 @@ const (
 	// fmtRID is the format of the resource ID that has a virtual machine name
 	fmtRID = "/subscriptions/%s/resourcegroups/%s/providers/Microsoft.Compute/virtualMachines/%s"
 
+	// fmtVMSSRID is the format of the resource ID when VMSS is in uniform mode that just has the scaleset name
+	fmtVMSSRID = "/subscriptions/%s/resourcegroups/%s/providers/Microsoft.Compute/virtualMachineScaleSet/%s"
+
 	// fmtRIDWithUserAssignedIdentities is the format of the resource ID that has a user-assigned managed identity
 	fmtRIDWithUserAssignedIdentities = "/subscriptions/%s/resourcegroups/%s/providers/Microsoft.ManagedIdentity/userAssignedIdentities/%s"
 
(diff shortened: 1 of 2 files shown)


wilsonge:
cc'ing @thyton as you wrote the original code

path_login.go
```diff
@@ -618,9 +625,14 @@ func (c *additionalClaims) verifyVM(vmName string) error {
 }
 
 // verifyVMSS checks the additional claims in the token against
-// the provided vm_name field on login
+// the provided vm_name field on login. We have to check both the uniform and flexible
+// claim patterns.
```
helenfufu, line 629: Minor suggestion to provide some reasoning around why we try the flexible claim pattern first (which makes sense given it's the recommended and default orchestration option from Microsoft)

```suggestion
// the provided vm_name field on login. We have to check both the
// uniform and flexible claim patterns. Since flexible is recommended
// and default (https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-orchestration-modes), check it first.
// If flexible fails, fall back to uniform.
```

path_login.go
```diff
@@ -618,9 +625,14 @@ func (c *additionalClaims) verifyVM(vmName string) error {
...
 func (c *additionalClaims) verifyVMSS(vmssName string) error {
-	return c.verifyXMSClaims(fmtVMSSClaimPattern, "vmss_name", vmssName)
+	err := c.verifyXMSClaims(fmtVMSSFlexibleClaimPattern, "vmss_name", vmssName)
+	if err != nil {
+		return c.verifyXMSClaims(fmtVMSSUniformClaimPattern, "vmss_name", vmssName)
+	}
+	return err
```
helenfufu, line 635: Could we keep track of multiple errors so in the case that we fail both verification steps, both errors will be returned?

```suggestion
	var errs []error
	if err := c.verifyXMSClaims(fmtVMSSFlexibleClaimPattern, "vmss_name", vmssName); err != nil {
		errs = append(errs, fmt.Errorf("failed to verify flexible vmss claim: %w", err))

		if err := c.verifyXMSClaims(fmtVMSSUniformClaimPattern, "vmss_name", vmssName); err != nil {
			errs = append(errs, fmt.Errorf("failed to verify uniform vmss claim: %w", err))
		}
	}

	return errors.Join(errs...)
```

path_login_test.go
```diff
@@ -1103,7 +1103,7 @@ func TestLogin_BoundScaleSet(t *testing.T) {
 			},
 		},
 		{
-			// The VMSS in this case has user-assigned managed identities
+			// The Flexible VMSS in this case has user-assigned managed identities
```
helenfufu, line 1106: Tiny nit: Could we keep the Flexible and Uniform capitalization consistent across the test cases? (Indifferent between capital or lowercase, but would prefer they're consistent!)

helenfufu (commented):
Hi @wilsonge, thank you so much for your contribution! This is looking great. Had one main suggestion about keeping track of and propagating both errors in the case that both flexible and uniform verification fail, and a couple smaller comments about documentation.

heatherezell:
> cc'ing @thyton as you wrote the original code

Hello! As a note, the engineer who is helping to maintain the code may not be the original author. Please do not tag engineers directly. Thanks!

thyton:
Thank you very much for your contribution, @wilsonge!

PR #271, opened 2026-05-18
Add native AKS Workload Identity support to Vault Azure Auth

santoshhashicorp:
# Overview

## Summary

Add native AKS Workload Identity support to Vault Azure Auth.

## What Changed

- This change introduces the aks_wi authentication type, allowing Vault running on AKS to authenticate with Microsoft Entra ID using federated workload identity instead of static client secrets or managed identities. It also enables cross-tenant authentication by explicitly using the configured tenant and client IDs.

# Related Ticket
- [Vault #43953](https://hashicorp.atlassian.net/browse/VAULT-43953)

# Contributor Checklist
image

image

image

- [x] Backwards compatible

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

### DIFF
--- a/azure.go
+++ b/azure.go
@@ -43,6 +43,12 @@ const (
 	azurePublicCloudEnvName = "AZUREPUBLICCLOUD"
 	azureChinaCloudEnvName  = "AZURECHINACLOUD"
 	azureUSGovCloudEnvName  = "AZUREUSGOVERNMENTCLOUD"
+
+	aadErrFederatedCredentialNotReady = "AADSTS70021"
+	authTypeRootCreds = "root_creds"
+	authTypePluginWIF = "plugin_wif"
+	authTypeAKSWI     = "aks_wi"
+	authTypeMSI       = "msi"
 )
 
 type provider interface {
@@ -53,9 +59,6 @@ type provider interface {
 	MSGraphClient() (client.MSGraphClient, error)
 	ResourceClient(subscriptionID string) (client.ResourceClient, error)
 	ProvidersClient(subscriptionID string) (client.ProvidersClient, error)
-	// VerifyCredential probes the configured credential by requesting a token.
-	// For aks_wi this detects AADSTS70021 (federated credential not yet propagated)
-	// before any ARM call is attempted.
 	VerifyCredential(ctx context.Context) error
 }
 
@@ -247,17 +250,15 @@ func (p *azureProvider) ResourceClient(subscriptionID string) (client.ResourceCl
 
 // isFederatedCredentialNotReady returns true when an Azure AD token exchange
 // fails because the federated identity credential has not yet propagated.
-// Azure AD returns AADSTS70021 in this window (typically up to ~60 seconds
-// after the federated credential is created).
 func isFederatedCredentialNotReady(err error) bool {
-	return err != nil && strings.Contains(err.Error(), "AADSTS70021")
+	return err != nil && strings.Contains(err.Error(), aadErrFederatedCredentialNotReady)
 }
 
 // VerifyCredential calls GetToken on the configured credential with the ARM
-// scope so that AADSTS70021 (federated credential not yet propagated) is
-// surfaced with a clear, actionable error at login time rather than being
-// buried inside a later ARM call failure — or going undetected entirely for
-// roles that use only bound_service_principal_ids (which make no ARM calls).
+// scope so that aadErrFederatedCredentialNotReady is surfaced with a clear,
+// actionable error at login time rather than being buried inside a later ARM
+// call failure — or going undetected for roles using only
+// bound_service_principal_ids (which make no ARM calls).
 func (p *azureProvider) VerifyCredential(ctx context.Context) error {
 	cred, err := p.getTokenCredential()
 	if err != nil {
@@ -269,7 +270,8 @@ func (p *azureProvider) VerifyCredential(ctx context.Context) error {
 	if err != nil {
 		if isFederatedCredentialNotReady(err) {
 			return fmt.Errorf("aks_wi: Vault's federated identity credential has not propagated yet "+
-				"(AADSTS70021); wait ~60 seconds after creating the credential in Azure AD and retry login: %w", err)
+				"(%s); wait ~60 seconds after creating the credential in Azure AD and retry login: %w",
+				aadErrFederatedCredentialNotReady, err)
 		}
 		return fmt.Errorf("aks_wi: failed to acquire Azure access token: %w", err)
 	}
@@ -296,20 +298,15 @@ func (p *azureProvider) getClientOptions() *arm.ClientOptions {
 func (p *azureProvider) getTokenCredential() (azcore.TokenCredential, error) {
 	cloudOpts := azcore.ClientOptions{Cloud: p.settings.CloudConfig}
 
-	authType := p.settings.AuthType
-	if authType == "" {
-		authType = "auto"
-	}
-
-	switch authType {
-	case "root_creds":
+	switch p.settings.AuthType {
+	case authTypeRootCreds:
 		// Explicit client secret credential. client_secret must be configured.
 		if p.settings.ClientSecret == "" {
 			return nil, errors.New("auth_type 'root_creds' requires client_secret to be configured")
 		}
 		return newClientSecretCred(p.settings.TenantID, p.settings.ClientID, p.settings.ClientSecret, cloudOpts)
 
-	case "plugin_wif":
+	case authTypePluginWIF:
 		// Explicit Vault plugin workload identity federation. identity_token_audience must be set.
 		if p.settings.IdentityTokenAudience == "" {
 			return nil, errors.New("auth_type 'plugin_wif' requires identity_token_audience to be configured")
@@ -317,15 +314,15 @@ func (p *azureProvider) getTokenCredential() (azcore.TokenCredential, error) {
 		return newClientAssertionCred(p.settings.TenantID, p.settings.ClientID,
 			getAssertionFunc(p.logger, p.systemView, p.settings), cloudOpts)
 
-	case "aks_wi":
+	case authTypeAKSWI:
 		// Explicit AKS Workload Identity. The SDK reads AZURE_FEDERATED_TOKEN_FILE automatically.
 		return newWorkloadIdentityCred(p.settings.TenantID, p.settings.ClientID, cloudOpts)
 
-	case "msi":
+	case authTypeMSI:
 		// Explicit managed service identity (IMDS).
 		return newManagedIdentityCred(p.settings.ClientID, cloudOpts)
 
-	default: // "auto" — backward-compatible waterfall.
+	default: // Backward-compatible discovery flow.
 		if p.settings.ClientSecret != "" {
 			return newClientSecretCred(p.settings.TenantID, p.settings.ClientID, p.settings.ClientSecret, cloudOpts)
 		}
--- a/azure.go
+++ b/azure.go
@@ -45,10 +45,10 @@ const (
 	azureUSGovCloudEnvName  = "AZUREUSGOVERNMENTCLOUD"
 
 	aadErrFederatedCredentialNotReady = "AADSTS70021"
-	authTypeRootCreds = "root_creds"
-	authTypePluginWIF = "plugin_wif"
-	authTypeAKSWI     = "aks_wi"
-	authTypeMSI       = "msi"
+	authTypeRootCreds                 = "root_creds"
+	authTypePluginWIF                 = "plugin_wif"
+	authTypeAKSWI                     = "aks_wi"
+	authTypeMSI                       = "msi"
 )
 
 type provider interface {
--- a/azure.go
+++ b/azure.go
@@ -294,7 +294,7 @@ func (p *azureProvider) getClientOptions() *arm.ClientOptions {
 }
 
 func (p *azureProvider) getTokenCredential() (azcore.TokenCredential, error) {
-	clientCloudOpts := azcore.ClientOptions{Cloud: p.settings.CloudConfig}
+	cloudOpts := azcore.ClientOptions{Cloud: p.settings.CloudConfig}
 
 	authType := p.settings.AuthType
 	if authType == "" {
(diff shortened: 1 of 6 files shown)


azure.go
```diff
@@ -289,20 +335,92 @@ func (p *azureProvider) getTokenCredential() (azcore.TokenCredential, error) {
...
+			ID:            azidentity.ClientID(p.settings.ClientID),
+		}
+		cred, err := azidentity.NewManagedIdentityCredential(options)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create managed identity token credential: %w", err)
+		}
+		return cred, nil
+
+	default: // "auto" — backward-compatible waterfall
+		if p.settings.ClientSecret != "" {
```
ankitsutharhashicorp, line 368: If auth type is `root_creds`, then Explicit client secret credential is used.
If auth type is `plugin_wif`,  then Explicit Vault plugin workload identity federation is used, provided identity_token_audience is set.

The same is also added under `default` as well. Is this expected that if no auth type is defined then also we follow same order under `default` which is already covered in earlier case?

path_config.go
```diff
@@ -208,6 +213,20 @@ func (b *azureAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Req
...
 
+	if authTypeRaw, ok := data.GetOk("auth_type"); ok {
+		authType := authTypeRaw.(string)
+		validAuthTypes := map[string]bool{
+			"root_creds": true,
+			"plugin_wif": true,
+			"aks_wi":     true,
+			"msi":        true,
+		}
+		if !validAuthTypes[authType] {
```
ankitsutharhashicorp, line 224: When authType is empty then it will return below error response. So in which case `default` should execute while fetching token in azure.go?

azure.go
santoshhashicorp: @ankitsutharhashicorp 
Yes, this is expected. The `default` flow follows the same authentication order, but the difference is that without `auth_type`, we don’t have explicit control over which authentication mechanism is used. The first valid configured mechanism is selected.

When `auth_type` is specified, only that authentication mechanism is used and its required parameters are validated.

path_config.go
santoshhashicorp: @ankitsutharhashicorp 
If auth_type is empty or not provided, GetOk("auth_type") returns false, so the validation block is skipped. In that case, config.AuthType remains unset and authentication proceeds through the default flow in azure.go.

azure.go
ankitsutharhashicorp: Ok, so in the default flow we evaluate in the order of `ClientSecret`,  `IdentityTokenAudience`, `AZURE_FEDERATED_TOKEN_FILE`. Is `IMDS` not falling into this order to use the Managed Identity attached to the hosting resource?

santoshhashicorp: @ankitsutharhashicorp 
Yes, that’s correct. In the default flow, we evaluate ClientSecret, IdentityTokenAudience, and AZURE_FEDERATED_TOKEN_FILE first. We intentionally give priority to the AKS workload identity mechanisms over IMDS. If none of these are configured, we then fall back to IMDS to use the Managed Identity attached to the hosting resource.

azure.go
```diff
@@ -261,21 +296,32 @@ func (p *azureProvider) getClientOptions() *arm.ClientOptions {
 func (p *azureProvider) getTokenCredential() (azcore.TokenCredential, error) {
 	clientCloudOpts := azcore.ClientOptions{Cloud: p.settings.CloudConfig}
 
-	if p.settings.ClientSecret != "" {
+	authType := p.settings.AuthType
+	if authType == "" {
+		authType = "auto"
```
roh-a: [Doubt] Is there a reason we are setting the value of `authType` to `auto`? Can we leave it empty?

path_config.go
```diff
@@ -208,6 +213,20 @@ func (b *azureAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Req
 		config.ClientSecret = clientSecret.(string)
 	}
 
+	if authTypeRaw, ok := data.GetOk("auth_type"); ok {
```
roh-a: Currently do we have a way to unset the `auth_type` once it has been set? In subsequent config writes, if `auth_type` is explicitly set to an empty string, should we allow it to be unset?

azure.go
```diff
@@ -289,20 +335,92 @@ func (p *azureProvider) getTokenCredential() (azcore.TokenCredential, error) {
...
+			cred, err := azidentity.NewClientSecretCredential(p.settings.TenantID, p.settings.ClientID,
+				p.settings.ClientSecret, options)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create client secret token credential: %w", err)
+			}
+			return cred, nil
+		}
 
-	return cred, nil
+		if p.settings.IdentityTokenAudience != "" {
```
stuti-sr, line 329: The "auto" mode waterfall duplicates credential creation logic from the explicit cases.  
Lines 368-378 duplicate 305-318 (root_creds)
Lines 380-394 duplicate 320-338 (plugin_wif)
Lines 400-410 duplicate 340-353 (aks_wi)
Lines 414-422 duplicate 355-365 (msi)
Is there a reason for duplicating the code?

azure.go
```diff
@@ -289,20 +335,92 @@ func (p *azureProvider) getTokenCredential() (azcore.TokenCredential, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("failed to create client assertion credential: %w", err)
+			}
+			return cred, nil
+		}
+
+		// The presence of AZURE_FEDERATED_TOKEN_FILE indicates an AKS workload identity context.
+		// Use WorkloadIdentityCredential with explicit ClientID/TenantID from plugin config to
+		// enable cross-tenant ARM calls (avoids InvalidAuthenticationTokenTenant errors).
+		if os.Getenv("AZURE_FEDERATED_TOKEN_FILE") != "" {
```
stuti-sr, line 333: Should we also verify the credential when AKS Workload Identity is auto-detected? 
Currently, WorkloadIdentityCredential is used when AZURE_FEDERATED_TOKEN_FILE is present, but verification only happens for auth_type == "aks_wi". Is this intentional?
path_login.go : line(223)

santoshhashicorp: @stuti-sr Thanks for the suggestion.
Added a check for auth_type == "aks_wi" and when auth_type is empty then we verify whether AZURE_FEDERATED_TOKEN_FILE is present before calling VerifyCredential.
https://github.com/hashicorp/vault-plugin-auth-azure/blob/ac25eaad0424c933be452c1ee59fa7c595ab2865/path_login.go#L228-L229

azure.go
santoshhashicorp: @stuti-sr Thanks for the suggestion. Removed the duplicate code and created reusable functions.

stuti-sr (dismissed):
LGTM!!

path_config.go
```diff
@@ -208,6 +213,20 @@ func (b *azureAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Req
...
+		authType := authTypeRaw.(string)
+		validAuthTypes := map[string]bool{
+			"root_creds": true,
+			"plugin_wif": true,
+			"aks_wi":     true,
+			"msi":        true,
+		}
```
roh-a: [Nit] Does it make sense to create string constants for different auth type values as they are re-used in multiple places?

azure.go
```diff
@@ -241,6 +245,37 @@ func (p *azureProvider) ResourceClient(subscriptionID string) (client.ResourceCl
...
+	if err != nil {
+		return fmt.Errorf("failed to build Azure credential: %w", err)
+	}
+	_, err = cred.GetToken(ctx, policy.TokenRequestOptions{
+		Scopes: []string{p.settings.Resource + "/.default"},
+	})
+	if err != nil {
+		if isFederatedCredentialNotReady(err) {
+			return fmt.Errorf("aks_wi: Vault's federated identity credential has not propagated yet "+
+				"(AADSTS70021); wait ~60 seconds after creating the credential in Azure AD and retry login: %w", err)
```
roh-a: [Nit] Does it make sense to create a constant for the error code as it is used in multiple places? `AADSTS70021`

roh-a (commented):
Good work! Left some minor comments.

path_config.go
```diff
@@ -81,6 +81,10 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 				Description: "The initial amount of delay to use before retrying an operation, increasing exponentially.",
 				Required:    false,
 			},
+			"auth_type": {
+				Type:        framework.TypeString,
```
roh-a: [Nit] Should we add `Required:    false` for documentation.

santoshhashicorp: Thanks for the suggestion. I've updated the code accordingly.

azure.go
santoshhashicorp: Thanks for the suggestion. I've updated the code accordingly.

path_config.go
santoshhashicorp: Thanks for the suggestion. I've updated the code accordingly.

azure.go
santoshhashicorp: Yes, that makes sense. I've updated the code accordingly.

path_config.go
santoshhashicorp: @roh-a Thanks for highlighting this issue. 
If **auth_type** is provided during a config, its value should be preserved across subsequent config updates/reads until explicitly changed. If auth_type is explicitly set to an empty value, it should overwrite and replace the existing stored value accordingly and **auth_type** should always be present in the read config response.

Results:
image
image

roh-a (dismissed):
Looks good to me!

ankitsutharhashicorp (dismissed):
LGTM!

path_config_test.go
```diff
@@ -69,6 +71,86 @@ func TestConfig(t *testing.T) {
 				"disable_automated_rotation": false,
 			},
 		},
+		{
+			name: "auth_type aks_wi happy path",
+			config: map[string]interface{}{
+				"resource":  "resource",
+				"tenant_id": "tid",
+				"client_id": "my-managed-identity-client-id",
+				"auth_type": "aks_wi",
```
stuti-sr, line 80: Is there a test for auth_type=aks_wi without client_id set?

azure.go
```diff
@@ -241,6 +248,36 @@ func (p *azureProvider) ResourceClient(subscriptionID string) (client.ResourceCl
...
+	_, err = cred.GetToken(ctx, policy.TokenRequestOptions{
+		Scopes: []string{p.settings.Resource + "/.default"},
+	})
+	if err != nil {
+		if isFederatedCredentialNotReady(err) {
+			return fmt.Errorf("aks_wi: Vault's federated identity credential has not propagated yet "+
+				"(%s); wait ~60 seconds after creating the credential in Azure AD and retry login: %w",
+				aadErrFederatedCredentialNotReady, err)
+		}
+		return fmt.Errorf("aks_wi: failed to acquire Azure access token: %w", err)
```
stuti-sr: This method is called for both config.AuthType == authTypeAKSWI and the auto-detect path (config.AuthType == "" && AZURE_FEDERATED_TOKEN_FILE != ""). Is this intentional to provide error as fmt.Errorf("aks_wi:....") for both scenarios?

path_config.go
```diff
@@ -244,6 +274,13 @@ func (b *azureAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Req
 		return logical.ErrorResponse("only one of 'client_secret' or 'identity_token_audience' can be set"), nil
 	}
 
+	if config.AuthType == authTypeRootCreds && config.ClientSecret == "" {
```
stuti-sr, line 277: should there also be a check to validate for authTypeAKSWI & config.ClientID?

stuti-sr (commented):
provided comments

azure.go
santoshhashicorp: Thanks for the suggestion. I’'ve updated it to **azure_workload_identity** to keep the error message generic across both scenarios.

path_config_test.go
santoshhashicorp: Added test for auth_type=aks_wi without client_id now.

path_config.go
santoshhashicorp: No check is needed because client_id is intentionally optional at config-write time for aks_wi, as it can be injected by the AKS Workload Identity via AZURE_CLIENT_ID.

path_login_test.go
```diff
@@ -180,6 +180,123 @@ func TestLogin(t *testing.T) {
...
+// propagated yet and Azure AD returns AADSTS70021), login is rejected before
+// JWT verification with the propagation error surfaced to the caller.
+func TestLogin_AKS_WI_CredentialNotReady(t *testing.T) {
+	b, s := getTestBackend(t)
+
+	configData := map[string]interface{}{
+		"tenant_id": "test-tenant-id",
+		"resource":  "https://management.azure.com/",
+		"client_id": "test-managed-identity-client-id",
+		"auth_type": "aks_wi",
```
stuti-sr, line 263: I see test for config.AuthType == aks_wi. 
Is there any test for config.AuthType == "" && AZURE_FEDERATED_TOKEN_FILE != ""?

path_config.go
```diff
@@ -208,6 +219,25 @@ func (b *azureAuthBackend) pathConfigWrite(ctx context.Context, req *logical.Req
 		config.ClientSecret = clientSecret.(string)
 	}
 
+	if _, sent := data.Raw["auth_type"]; sent {
+		// auth_type was explicitly included in the request — always overwrite, even if empty.
+		// omitting auth_type preserves the stored value.
+		authType := data.Get("auth_type").(string)
+		if authType != "" {
```
stuti-sr, line 226: What would happen when explicitly setting auth_type="" when credentials such as client_secret are already present from a previous root_creds configuration?

path_config.go
stuti-sr: What will happen when config.ClientID == "" && os.Getenv("AZURE_CLIENT_ID") == "" ?

azure.go
```diff
@@ -241,6 +248,36 @@ func (p *azureProvider) ResourceClient(subscriptionID string) (client.ResourceCl
...
+	return err != nil && strings.Contains(err.Error(), aadErrFederatedCredentialNotReady)
+}
+
+// VerifyCredential calls GetToken on the configured credential with the ARM
+// scope so that aadErrFederatedCredentialNotReady is surfaced with a clear,
+// actionable error at login time rather than being buried inside a later ARM
+// call failure — or going undetected for roles using only
+// bound_service_principal_ids (which make no ARM calls).
+func (p *azureProvider) VerifyCredential(ctx context.Context) error {
+	cred, err := p.getTokenCredential()
```
stuti-sr, line 263: VerifyCredential has no timeout of its own. Is this intentional ?

stuti-sr (commented):
provided comments

azure.go
santoshhashicorp: Yes, intentional. VerifyCredential uses the Vault request context, which already carries the request timeout.

path_config.go
santoshhashicorp: For AKS Workload Identity, `AZURE_CLIENT_ID` is injected into the Vault pod by the Workload Identity webhook, so `os.Getenv("AZURE_CLIENT_ID")` should not be empty at runtime.

path_config.go
santoshhashicorp: If auth_type is provided during a config, its value should be preserved across subsequent config updates/reads until explicitly changed. If auth_type is explicitly set to an empty value, it should overwrite and replace the existing stored value accordingly.

The same has been addressed in this comment:
https://github.com/hashicorp/vault-plugin-auth-azure/pull/271#discussion_r3767311855

path_login_test.go
santoshhashicorp: Added TestLogin_AutoDetect_FederatedTokenFile.

roh-a (approved):

stuti-sr (approved):
LGTM!!

PR #272, opened 2026-08-18
Added missing OpenAPI metadata to the API endpoints

AnujPatnaik1:
# Overview
Who the change affects or is for (stakeholders)?
The change affects mainly people who use the Vault's UI. With the added fields, they can better understand the endpoints in the azure auth plugin.

What is the change?
I added the missing operation summary, path descriptions, and proper typed response schema for the azure auth plugin. I edited path_config.go, path_login.go, path_role.go, and path_rotate_root.go

Why is the change needed?
The change is needed because many of the plugins in vault were missing path descriptions, operation summaries, and typed response schemas. Many of the response schemas had a bare 200 Description OK response, which is not sufficient.

How does this change affect the user experience (if at all)?
When someone uses the Vault UI, they are able to understand the vault plugins better because they have a detailed description of what the endpoint does and meaningful response schema. With the response schema, they can understand what the endpoint returns.

**Design**
I developed an AI skill that automatically adds the HelpSynopsis, Summary, and Response Fields. The Help Synopsis field helps with the path descriptions, and the Help Description and path name are used to come up with the Help Synopsis field. The path name, Operation Prefix, and Operation Suffix are used to come up with the operation summary. For the response schema, the AI skill is able to detect what data types are being added to the response data map, and the skill generates a response schema with those data types. For methods that return a nil, nil, a 204 response code is added.

### DIFF
--- a/path_config.go
+++ b/path_config.go
@@ -141,7 +141,7 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 								Description: "The maximum number of attempts a failed operation will be retried.",
 							},
 							"root_password_expiration_date": {
-								Type:        framework.TypeString,
+								Type:        framework.TypeTime,
 								Description: "The expiration date of the root password.",
 							},
 							"identity_token_ttl": {
--- a/path_config.go
+++ b/path_config.go
@@ -129,11 +129,11 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 								Description: "The TTL of the root password in Azure.",
 							},
 							"retry_delay": {
-								Type:        framework.TypeDurationSecond,
+								Type:        framework.TypeSignedDurationSecond,
 								Description: "The initial amount of delay to use before retrying an operation.",
 							},
 							"max_retry_delay": {
-								Type:        framework.TypeDurationSecond,
+								Type:        framework.TypeSignedDurationSecond,
 								Description: "The maximum delay allowed before retrying an operation.",
 							},
 							"max_retries": {
@@ -157,11 +157,11 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 								Description: "The schedule for automated root credential rotation.",
 							},
 							"rotation_window": {
-								Type:        framework.TypeInt,
+								Type:        framework.TypeDurationSecond,
 								Description: "The maximum time allowed for a rotation to complete.",
 							},
 							"rotation_period": {
-								Type:        framework.TypeInt,
+								Type:        framework.TypeDurationSecond,
 								Description: "The period for automated root credential rotation.",
 							},
 							"disable_automated_rotation": {
--- a/path_config.go
+++ b/path_config.go
@@ -145,7 +145,7 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 								Description: "The expiration date of the root password.",
 							},
 							"identity_token_ttl": {
-								Type:        framework.TypeDurationSecond,
+								Type:        framework.TypeInt,
 								Description: "Time-to-live of plugin identity tokens.",
 							},
 							"identity_token_audience": {
@@ -157,11 +157,11 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 								Description: "The schedule for automated root credential rotation.",
 							},
 							"rotation_window": {
-								Type:        framework.TypeDurationSecond,
+								Type:        framework.TypeInt,
 								Description: "The maximum time allowed for a rotation to complete.",
 							},
 							"rotation_period": {
-								Type:        framework.TypeDurationSecond,
+								Type:        framework.TypeInt,
 								Description: "The period for automated root credential rotation.",
 							},
 							"disable_automated_rotation": {
--- a/path_config.go
+++ b/path_config.go
@@ -184,7 +184,7 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 				},
 				ForwardPerformanceSecondary: true,
 				ForwardPerformanceStandby:   true,
-				Summary: "Configure the Azure authentication backend.",
+				Summary:                     "Configure the Azure authentication backend.",
 				Responses: map[int][]framework.Response{
 					204: {{Description: "No Content"}},
 				},
@@ -197,7 +197,7 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 				},
 				ForwardPerformanceSecondary: true,
 				ForwardPerformanceStandby:   true,
-				Summary: "Configure the Azure authentication backend.",
+				Summary:                     "Configure the Azure authentication backend.",
 				Responses: map[int][]framework.Response{
 					204: {{Description: "No Content"}},
 				},
--- a/path_config.go
+++ b/path_config.go
@@ -99,6 +99,82 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
 					OperationVerb:   "read",
 					OperationSuffix: "auth-configuration",
 				},
+				Summary: "Return the current Azure authentication backend configuration.",
+				Responses: map[int][]framework.Response{
+					200: {{
+						Description: "OK",
+						Fields: map[string]*framework.FieldSchema{
+							"tenant_id": {
+								Type:        framework.TypeString,
+								Description: "The tenant id for the Azure Active Directory.",
+							},
+							"resource": {
+								Type:        framework.TypeString,
+								Description: "The resource URL for the vault application in Azure Active Directory.",
+							},
+							"environment": {
+								Type:        framework.TypeString,
+								Description: "The Azure environment name.",
+							},
+							"client_id": {
+								Type:        framework.TypeString,
+								Description: "The OAuth2 client id to connection to Azure.",
+							},
+							"auth_type": {
+								Type:        framework.TypeString,
+								Description: "The authentication type used to connect to Azure.",
+							},
+							"root_password_ttl": {
+								Type:        framework.TypeDurationSecond,
+								Description: "The TTL of the root password in Azure.",
+							},
+							"retry_delay": {
+								Type:        framework.TypeDurationSecond,
+								Description: "The initial amount of delay to use before retrying an operation.",
+							},
+							"max_retry_delay": {
+								Type:        framework.TypeDurationSecond,
+								Description: "The maximum delay allowed before retrying an operation.",
+							},
+							"max_retries": {
+								Type:        framework.TypeInt,
+								Description: "The maximum number of attempts a failed operation will be retried.",
+							},
+							"root_password_expiration_date": {
+								Type:        framework.TypeString,
+								Description: "The expiration date of the root password.",
+							},
+							"identity_token_ttl": {
+								Type:        framework.TypeDurationSecond,
+								Description: "Time-to-live of plugin identity tokens.",
+							},
+							"identity_token_audience": {
+								Type:        framework.TypeString,
+								Description: "Audience of plugin identity tokens.",
+							},
+							"rotation_schedule": {
+								Type:        framework.TypeString,
+								Description: "The schedule for automated root credential rotation.",
+							},
+							"rotation_window": {
+								Type:        framework.TypeDurationSecond,
+								Description: "The maximum time allowed for a rotation to complete.",
+							},
+							"rotation_period": {
+								Type:        framework.TypeDurationSecond,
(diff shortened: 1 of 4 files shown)


path_config.go
```diff
@@ -99,6 +99,82 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
...
+							"max_retries": {
+								Type:        framework.TypeInt,
+								Description: "The maximum number of attempts a failed operation will be retried.",
+							},
+							"root_password_expiration_date": {
+								Type:        framework.TypeString,
+								Description: "The expiration date of the root password.",
+							},
+							"identity_token_ttl": {
+								Type:        framework.TypeDurationSecond,
```
fairclothjm: is this the correct type?

path_login.go
```diff
@@ -122,12 +122,70 @@ func pathLogin(b *azureAuthBackend) *framework.Path {
 		Operations: map[logical.Operation]framework.OperationHandler{
 			logical.UpdateOperation: &framework.PathOperation{
 				Callback: b.pathLogin,
+				Summary:  "Authenticate to Vault using Azure credentials.",
+				Responses: map[int][]framework.Response{
+					200: {{
+						Description: "OK",
+						Fields: map[string]*framework.FieldSchema{
```
rana, line 129: Are Fields necessary in this case? Those are Vault Core and not plugin?

See [path_login.go#L38-L44](https://github.com/hashicorp/vault/blob/main/builtin/credential/approle/path_login.go#L38-L44) for comparison.

path_role.go
```diff
@@ -98,15 +110,115 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
 				},
 				logical.ReadOperation: &framework.PathOperation{
 					Callback: b.pathRoleRead,
+					Summary:  "Return the properties of an Azure authentication role.",
+					Responses: map[int][]framework.Response{
+						200: {{
+							Description: "OK",
+							Fields: map[string]*framework.FieldSchema{
+								"bound_service_principal_ids": {
+									Type:        framework.TypeSlice,
```
rana: Consider changing the type to `TypeCommaStringSlice`? I believe the return type is a slice of strings. See [L84-L85](https://github.com/hashicorp/vault-plugin-auth-azure/blob/c968e02a48a4b5fe532e005efdf7cce8e76c2036/path_role.go#L84-L85).

```suggestion
									Type:        framework. TypeCommaStringSlice,
```

I believe the same change applies to other fields as well: each `bound_*`, `token_bound_cidrs`, `token_policies`, and `policies`.

path_role.go
```diff
@@ -28,6 +28,18 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+						200: {{
+							Description: "OK",
+							Fields: map[string]*framework.FieldSchema{
+								"keys": {
+									Type:        framework.TypeSlice,
+									Description: "List of role names.",
+								},
+							},
+						}},
+					},
```
rana: Consider removing the `Responses` block for `list` because it would automatically be written? See [openapi.go#L538-L539](https://github.com/hashicorp/vault/blob/main/sdk/framework/openapi.go#L538-L539).

path_config.go
```diff
@@ -99,6 +99,82 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
...
+							"identity_token_audience": {
+								Type:        framework.TypeString,
+								Description: "Audience of plugin identity tokens.",
+							},
+							"rotation_schedule": {
+								Type:        framework.TypeString,
+								Description: "The schedule for automated root credential rotation.",
+							},
+							"rotation_window": {
+								Type:        framework.TypeInt,
```
rana: Consider changing type to `TypeDurationSecond`?  Is there a type mismatch? See:
- [automatedrotationutil/fields.go#L387-L392](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L387-L392)
- [automatedrotationutil/fields.go#L394-L399](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L394-L399)
- [automatedrotationutil/fields.go#L283-L284](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L283-L284)

```suggestion
								Type:        framework.TypeDurationSecond,
```

path_config.go
```diff
@@ -99,6 +99,82 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
...
+								Description: "The maximum time allowed for a rotation to complete.",
+							},
+							"rotation_period": {
+								Type:        framework.TypeInt,
```
rana: Consider changing type to `TypeDurationSecond`?  Is there a type mismatch? See:
- [automatedrotationutil/fields.go#L387-L392](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L387-L392)
- [automatedrotationutil/fields.go#L394-L399](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L394-L399)
- [automatedrotationutil/fields.go#L283-L284](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L283-L284)

```suggestion
								Type:        framework.TypeDurationSecond,
```

path_role.go
```diff
@@ -98,15 +110,115 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+								"bound_scale_sets": {
+									Type:        framework.TypeSlice,
+									Description: "List of scale sets that login is restricted to.",
+								},
+								"token_bound_cidrs": {
+									Type:        framework.TypeSlice,
+									Description: "List of CIDR blocks that tokens are restricted to.",
+								},
+								"token_explicit_max_ttl": {
+									Type:        framework.TypeInt,
```
rana: Consider using `TypeInt64` for explicitness?

Similar change for related fields `token_max_ttl`, `token_period`, `token_ttl`?

path_role.go
```diff
@@ -98,15 +110,115 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+									Description: "The incremental lifetime for generated tokens.",
+								},
+								"token_num_uses": {
+									Type:        framework.TypeInt,
+									Description: "The maximum number of uses for the token.",
+								},
+								"policies": {
+									Type:        framework.TypeSlice,
+									Description: "Deprecated: use token_policies instead.",
```
rana: Consider adding `Deprecated: true` for any depreciated fields?

rana (commented):
Hi Anuj,

Thank you for your efforts.
Comments inline. 
Similar suggested changes to the `vault-enterprise` PR.

path_login.go
AnujPatnaik1: The fields are necessary because that's what the UI team wants in terms of a response schema.

path_role.go
AnujPatnaik1: For the bound_service_principal_ids, they return it as "BoundServicePrincipalIDs []string `json:"bound_service_principal_ids"". That's why it's TypeSlice and not TypeCommaStringSlice. Also, looking at the code, I think TypeCommaStringSlice is the datatype for inputs, but for the outputs it should be a TypeSlice. The TypeCommaStringSlice is for separating the input by a comma, but for the output it should be returned as a TypeSlice.

path_login.go
rana: Discussed on call to leave as-is.

path_role.go
rana: Since `BoundServicePrincipalIDs []string` returns a slice of strings, would it make sense for OpenAPI to return a slice of strings, instead of a slice of objects?

My understanding is that TypeSlice emits slice of objects instead of strings.

Looking at the OpenAPI generator code in the SDK: [sdk/framework/openapi.go L1186-L1191](https://github.com/hashicorp/vault/blob/main/sdk/framework/openapi.go#L1186-L1191):

```go
case TypeSlice:
    ret.baseType = "array"
    ret.items = "object"          // ← wrong for []string
case TypeStringSlice, TypeCommaStringSlice:
    ret.baseType = "array"
    ret.items = "string"          // ← correct for []string
```

Here are a couple examples of strings slices in Vault Enterprise:
- `TypeStringSlice` - [pki/path_export_issuer_ent.go#L111](https://github.com/hashicorp/vault-enterprise/blob/7f1cff19cf0bbe7f8f30bb4806638cabe2330c96/builtin/logical/pki/path_export_issuer_ent.go#L111)
- `TypeCommaStringSlice` - [pki/path_config_urls.go#L28-L29](https://github.com/hashicorp/vault-enterprise/blob/7f1cff19cf0bbe7f8f30bb4806638cabe2330c96/builtin/logical/pki/path_config_urls.go#L28-L29)

Doing a `FindAll = StringSlice` I see that there are a mix of `Comma` and no comma, making either pattern valid for output.

What do you think?

path_role.go
rana: Additions look good.

path_role.go
rana: Changes look good.

path_config.go
```diff
@@ -99,6 +99,82 @@ func pathConfig(b *azureAuthBackend) *framework.Path {
...
+							"max_retry_delay": {
+								Type:        framework.TypeSignedDurationSecond,
+								Description: "The maximum delay allowed before retrying an operation.",
+							},
+							"max_retries": {
+								Type:        framework.TypeInt,
+								Description: "The maximum number of attempts a failed operation will be retried.",
+							},
+							"root_password_expiration_date": {
+								Type:        framework.TypeString,
```
rana: Is this one `TypeTime`?
```suggestion
								Type:        framework.TypeTime,
```

path_role.go
```diff
@@ -98,15 +99,120 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+									Type:        framework.TypeInt64,
+									Description: "The incremental lifetime for generated tokens.",
+								},
+								"token_num_uses": {
+									Type:        framework.TypeInt,
+									Description: "The maximum number of uses for the token.",
+								},
+								"policies": {
+									Type:        framework.TypeCommaStringSlice,
+									Description: "Deprecated: use token_policies instead.",
```
rana: How about calling the helper function for consistency?
```suggestion
									Description: tokenutil.DeprecationText("token_policies"),
```

path_role.go
```diff
@@ -98,15 +99,120 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+									Deprecated:  true,
+								},
+								"ttl": {
+									Type:        framework.TypeInt64,
+									Description: "Deprecated: use token_ttl instead.",
```
rana: How about calling the helper function for consistency?
```suggestion
									Description: tokenutil.DeprecationText("token_ttl"),
```

path_role.go
```diff
@@ -98,15 +99,120 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+									Deprecated:  true,
+								},
+								"max_ttl": {
+									Type:        framework.TypeInt64,
+									Description: "Deprecated: use token_max_ttl instead.",
```
rana: How about calling the helper function for consistency?
```suggestion
									Description: tokenutil.DeprecationText("token_max_ttl"),
```

path_role.go
```diff
@@ -98,15 +99,120 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+									Deprecated:  true,
+								},
+								"period": {
+									Type:        framework.TypeInt64,
+									Description: "Deprecated: use token_period instead.",
```
rana: How about calling the helper function for consistency?
```suggestion
									Description: tokenutil.DeprecationText("token_period"),
```

path_role.go
```diff
@@ -98,15 +99,120 @@ func pathsRole(b *azureAuthBackend) []*framework.Path {
...
+									Deprecated:  true,
+								},
+								"num_uses": {
+									Type:        framework.TypeInt,
+									Description: "Deprecated: use token_num_uses instead.",
```
rana: How about calling the helper function for consistency?
```suggestion
									Description: tokenutil.DeprecationText("token_num_uses"),
```

rana (commented):
Hi Anuj,

Changes look good. 
I spotted a few more minor changes and made some suggestions.

rana (approved):
Changes look good.
Thank you!

